PR #37, opened 2021-05-24
Adding CDN Dashboard

arman-rye:
Adding a monitoring dashboard for Cloud CDN


arman-rye:
@googlebot I signed it!

jieshi1010:
The number on the Total Egress chart doesn't change when I change the time range. How can I achieve this?

PR #45, opened 2021-07-15
Update metric types for Headless vmare agent

schmikei:
These dashboards were clarified to be volatile so here is the update for them.  

The standard for the metrics that are written to by the GCVE agent now have changed. The latest version of the agent now reports to the namespace `external.googleapis.com/vmware/vcenter...`. 

This PR changes the metric types so that the visualizations will work with the latest version of the agent. 

Contention: 

image

Overview:

image

VM Performance (ignore the temporarily disabled metrics, chart definition is still valid with given descriptor):

image


xiangshen-dk:
Hi Keith,

Can you please also update the screenshots (and the README if needed) for the dashboards? (https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples#screenshots) 

Thanks!

PR #63, opened 2021-10-01
Add Cassandra dashboard

JonathanWamsley:
This pull request is contingent upon the [corresponding metrics receiver pull request](https://github.com/GoogleCloudPlatform/ops-agent/pull/220).


draffensperger:
Hi Jonathan! I'm an engineer at Google who works more on the frontend UI side of Cloud Monitoring and have been looking at some of these sample dashboards as they come by.

As a quick question - what does the Latency Count metric mean? Is that the same as an operation count?

And would it make sense to specify units for the latency charts? The numbers in the sample dashboard looked large (maybe microseconds or something?) There is a `unitOverride` field that you can put in the chart JSON (https://cloud.google.com/monitoring/api/ref_v3/rest/v1/projects.dashboards#timeseriesquery) that could be e.g. "us" (for microseconds) or "ns" for nanoseconds per https://cloud.google.com/monitoring/mql/reference#units-of-measure

djaglowski:
@draffensperger, good catch. You're right about those latency counts actually being operation counts. I can see why the original author of the data model made the mistake though, as Cassandra defines these counts as part of a "Latency" MBean. My team has submitted a [PR to fix the receiver](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/111#discussion_r726728215).

Regarding units on the other latency charts, we'll plan on overriding them in this case. Apparently there is a known issue with Google's exporter that is causing the loss of units.

PR #74, opened 2022-01-14
Correct screenshots for memcached & apache

pz-google:
Sample dashboards requires the filename of the dashboard spec to be the same with the filename of the screenshots.


xiangshen-dk (approved):
LGTM

PR #91, opened 2022-02-14
Fixes units on Solr Average Time charts

StefanKurek:


qingling128:
@xiangshen-dk Could you help with this one as well?

PR #117, opened 2022-04-07
Add Varnish Dashboard

JonathanWamsley:
This depends on the ops-agent varnish metric receiver [pr](https://github.com/GoogleCloudPlatform/ops-agent/pull/512) and the varnish log receiver


dashboards/varnish/metadata.yaml
```diff
@@ -0,0 +1,11 @@
+category: Varnish
```
EvanSimpson, line 1: I'll note that the format for `metadata.yaml` is still being finalized so there's expectation for it to be included as of now. Feel free to leave or revert, I'll make any necessary adjustments later and outline the expected format in the contribution docs when it's ready.

EvanSimpson (approved):

PR #133, opened 2022-04-29
Add Active Directory Domain Services dashboards

BinaryFissionGames:
Adds dashboards for the Active Directory Domain Services ops-agent integration.

Dependent on https://github.com/GoogleCloudPlatform/ops-agent/pull/563


dashboards/active-directory-ds/README.md
```diff
@@ -0,0 +1,11 @@
+- These dashboards are based on Google's [Ops Agent](https://cloud.google.com/stackdriver/docs/solutions/agents/ops-agent).
```
EvanSimpson, line 1: Can you add the top level and "Notes" headers here, as the other Readme files do?
e.g. https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/blob/master/dashboards/activemq/README.md

dashboards/active-directory-ds/metadata.yaml
```diff
@@ -0,0 +1,14 @@
+category: Active Directory
```
EvanSimpson, line 1: I'll note that the format for `metadata.yaml` is still being finalized so there's expectation for it to be included as of now. Feel free to leave or revert, I'll make any necessary adjustments later and outline the expected format in the contribution docs when it's ready.

dashboards/active-directory-ds/README.md
BinaryFissionGames: Added these headers, good catch.

dashboards/active-directory-ds/metadata.yaml
BinaryFissionGames: I've reverted the file for now.

EvanSimpson (approved):

PR #150, opened 2022-05-16
Alert policies for MSSQL ops agent (DRAFT)

schmikei:
Adds policies for a new ops agent integration located in [this PR](https://github.com/GoogleCloudPlatform/ops-agent/pull/554).

The alerts intended to be added:
- User Connections near limit - if the amount of connections is within 10% of the maximum connections default being 32,767, then that could be cause to look for ways to increase performance and load management
- High Page Split Rates - if ‘page.split.rate’ is spiking above a user defined threshold, it could show if the fill factor needs to increase. Excess page splitting can cause excessive disk I/O and contribute to slow performance.
- High Lock Wait Rates - if ‘lock.wait.rate’ is above a user defined value it shows that your resources are being blocked waiting for locks to be lifted slowing down the server.

### DIFF
--- a/dashboards/mssql-server/alerts/connections-near-limit.json
+++ b/dashboards/mssql-server/alerts/connections-near-limit.json
@@ -1,0 +1,38 @@
+{
+    "displayName": "MSSQL - Connections Near Limit",
+    "documentation": {
+        "content": "If the amount of connections is within 10% of the maximum connections, default being 32,767 (threshold is defaulted to 29490), then that could be cause to look for ways to increase performance and load management.",
+        "mimeType": "text/markdown"
+    },
+    "userLabels": {},
+    "conditions": [
+        {
+            "displayName": "MSSQL - User Connections Near Limit",
+            "conditionThreshold": {
+                "filter": "resource.type = \"gce_instance\" AND metric.type = \"workload.googleapis.com/sqlserver.user.connection.count\"",
+                "aggregations": [
+                    {
+                        "alignmentPeriod": "600s",
+                        "crossSeriesReducer": "REDUCE_NONE",
+                        "perSeriesAligner": "ALIGN_MEAN"
+                    }
+                ],
+                "comparison": "COMPARISON_GT",
+                "duration": "0s",
+                "trigger": {
+                    "count": 1
+                },
+                "thresholdValue": 29490
+            }
+        }
+    ],
+    "alertStrategy": {
+        "autoClose": "604800s",
+        "notificationPrompts": [
+            "OPENED"
+        ]
+    },
+    "combiner": "OR",
+    "enabled": true,
+    "notificationChannels": []
+}
--- a/dashboards/mssql-server/alerts/documentation.md
+++ b/dashboards/mssql-server/alerts/documentation.md
@@ -1,0 +1,14 @@
+
+# Alerts for MSSQL in the Ops Agent
+
+## Connections Near Limit
+
+If the amount of connections is within 10% of the maximum connections, then unexpected behavior can occur with connection establishment, queries and operations. By default the upper limit is `32,767` so the alert is specified for `29,490` connections. If limit is reached, it would be a potential cause of decreased performance of the instance.
+
+## High Page Split Rates
+
+If ‘page.split.rate’ is spiking above a user defined threshold (defaulted to `100 pages/sec`), it could show if the fill factor needs to increase. Excess page splitting can cause excessive disk I/O and contribute to slow performance of the instance.
+
+## High Lock Wait Rates
+
+If ‘lock.wait.rate’ is above a user defined value (defaulted to `3 requests/sec`) it shows that your resources are being blocked waiting for locks to be lifted slowing down the server.
--- a/dashboards/mssql-server/alerts/high-page-split-rate.json
+++ b/dashboards/mssql-server/alerts/high-page-split-rate.json
@@ -1,0 +1,38 @@
+{
+    "displayName": "MSSQL - HIgh Page Split Rates",
+    "documentation": {
+        "content": "If `page.split.rate` is spiking above a user defined threshold, it could show if the fill factor needs to increase. Excess page splitting can cause excessive disk I/O and contribute to slow performance.",
+        "mimeType": "text/markdown"
+    },
+    "userLabels": {},
+    "conditions": [
+        {
+            "displayName": "VM Instance - workload/sqlserver.page.split.rate",
+            "conditionThreshold": {
+                "filter": "resource.type = \"gce_instance\" AND metric.type = \"workload.googleapis.com/sqlserver.page.split.rate\"",
+                "aggregations": [
+                    {
+                        "alignmentPeriod": "300s",
+                        "crossSeriesReducer": "REDUCE_NONE",
+                        "perSeriesAligner": "ALIGN_MEAN"
+                    }
+                ],
+                "comparison": "COMPARISON_GT",
+                "duration": "0s",
+                "trigger": {
+                    "count": 1
+                },
+                "thresholdValue": 100
+            }
+        }
+    ],
+    "alertStrategy": {
+        "autoClose": "604800s",
+        "notificationPrompts": [
+            "OPENED"
+        ]
+    },
+    "combiner": "OR",
+    "enabled": true,
+    "notificationChannels": []
+}
--- a/dashboards/mssql-server/alerts/high-lock-wait-rates.json
+++ b/dashboards/mssql-server/alerts/high-lock-wait-rates.json
@@ -1,0 +1,38 @@
+{
+    "displayName": "MSSQL - High Lock Wait Rates",
+    "documentation": {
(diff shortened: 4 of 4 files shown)


dashboards/mssql-server/alerts/documentation.md
```diff
@@ -0,0 +1,14 @@
+
```
EvanSimpson, line 1: Can we rename this to `README.md` so GitHub will render it when viewing the folder?

dashboards/mssql-server/alerts/connections-near-limit.json
```diff
@@ -0,0 +1,38 @@
+{
```
EvanSimpson, line 1: Can you create a new top-level folder in this repo for alerts?
e.g.
/alerts/mssql-server/\*.json
instead of
/dashboards/mssql-server/alerts/\*.json

EvanSimpson (changes requested):

alerts/mssql-server/connections-near-limit.json
```diff
@@ -0,0 +1,38 @@
+{
+    "displayName": "MSSQL - Connections Near Limit",
+    "documentation": {
+        "content": "If the amount of connections is within 10% of the maximum connections, default being 32,767 (threshold is defaulted to 29490), then that could be cause to look for ways to increase performance and load management.",
+        "mimeType": "text/markdown"
+    },
+    "userLabels": {},
```
StevenYCChou, line 7:
```suggestion
```
Suggest removing the empty object here for simplicity.  This comment applies for other empty arrays in other files.

alerts/mssql-server/connections-near-limit.json
```diff
@@ -0,0 +1,38 @@
...
+    "conditions": [
+        {
+            "displayName": "MSSQL - User Connections Near Limit",
+            "conditionThreshold": {
```
StevenYCChou, line 11: Two questions -
1. when data stops arriving to Cloud Monitoring, what will you recommend to set `conditionThreshold.evaluationMissingData` [1] other than `EVALUATION_MISSING_DATA_UNSPECIFIED`? Can you set it explicitly? Here are the options:
   - `EVALUATION_MISSING_DATA_INACTIVE`: Missing data points treated as values that do not violate the policy condition.
   - `EVALUATION_MISSING_DATA_ACTIVE`: Missing data points treated as values that violate the policy condition
   - `EVALUATION_MISSING_DATA_NO_OP`: Do not evaluate the condition to any value if there is no data.
2. Follow-up of 1). If `conditionThreshold.evaluationMissingData` is `EVALUATION_MISSING_DATA_NO_OP`, do you recommend keeping the incident open for 7 days (as `alertStrategy.autoClose` was set to `604800s`)? Can it be shortened to a smaller duration? As a reference point, the minimum one can set on our UI is 30 mins (but it doesn't mean that it cannot be less than 30 mins).

This comment applies to the other two alerts as well.

[1] https://cloud.google.com/monitoring/api/ref_v3/rest/v3/projects.alertPolicies#metricthreshold

alerts/mssql-server/connections-near-limit.json
```diff
@@ -0,0 +1,38 @@
...
+    ],
+    "alertStrategy": {
+        "autoClose": "604800s",
+        "notificationPrompts": [
+            "OPENED"
+        ]
+    },
+    "combiner": "OR",
+    "enabled": true,
+    "notificationChannels": []
```
StevenYCChou, line 37:
```suggestion
```

Suggest removing the empty array here for simplicity. This comment applies for other empty arrays in other files.

alerts/mssql-server/connections-near-limit.json
schmikei: I did some thinking about this and I think my general thoughts are for these policies if metrics are absent it most likely is evident of an underlying issue of the Ops Agent. So I think specifying `evaluationMissingData: EVALUATION_MISSING_DATA_NO_OP` is a sensible approach.

Mostly because I believe the metrics are sourced from the `Windows Performance Counters` and is evident of another issue (most likely an underlying issue with the Ops Agent or host system) if the metrics end up not getting reported. 

However I do agree that 7 days for an autoclose period may be too long. I believe 1 hour is an appropriate value.

Do you have any similar or contradicting thoughts?

alerts/mssql-server/connections-near-limit.json
adeychman: Sorry for the back-and-forth, could we actually keep this empty array? User labels are a common addition made by our users and we want to make it easy for them to discover that it's possible to add tags to their alert policies.

dashboards/mssql-server/alerts/documentation.md
```diff
@@ -0,0 +1,14 @@
...
+# Alerts for MSSQL in the Ops Agent
+
+## Connections Near Limit
+
+If the amount of connections is within 10% of the maximum connections, then unexpected behavior can occur with connection establishment, queries and operations. By default the upper limit is `32,767` so the alert is specified for `29,490` connections. If limit is reached, it would be a potential cause of decreased performance of the instance.
+
+## High Page Split Rates
+
+If ‘page.split.rate’ is spiking above a user defined threshold (defaulted to `100 pages/sec`), it could show if the fill factor needs to increase. Excess page splitting can cause excessive disk I/O and contribute to slow performance of the instance.
```
adeychman: Suggested improvement: 

> If ‘page.split.rate’ is spiking above a user defined threshold (defaulted to `100 pages/sec`), the fill factor likely  needs to increase. Excess page splitting can cause excessive disk I/O and contribute to slow performance of the instance.

dashboards/mssql-server/alerts/high-page-split-rate.json
```diff
@@ -0,0 +1,38 @@
+{
+    "displayName": "MSSQL - HIgh Page Split Rates",
+    "documentation": {
+        "content": "If `page.split.rate` is spiking above a user defined threshold, it could show if the fill factor needs to increase. Excess page splitting can cause excessive disk I/O and contribute to slow performance.",
```
adeychman: We can keep consistent with the edit above:

> If ‘page.split.rate’ is spiking above a user defined threshold (defaulted to 100 pages/sec), the fill factor likely needs to increase. Excess page splitting can cause excessive disk I/O and contribute to slow performance of the instance.

alerts/mssql-server/connections-near-limit.json
StevenYCChou: TL;DR - I think In general I agree with you as how this PR stands, and I can mark it as approved as it is now. I have more considerations and I am going to explain below, but I think we probably can treat these additional considerations as a follow-up topic.

--

In general I agree with what you proposed with having `evaluationMissingData: EVALUATION_MISSING_DATA_NO_OP` and `autoClose` as 1h. I chatted with Alisa (@adeychman) and she seems fine with that. 

After I discussed with Alisa, I think about this thread more. 

There are two failure mode for these alert policies:
1) When MSSQL is operating, but Ops Agent and/or Cloud Monitoring API has issues. For example: a) Ops Agent is down; b) Cloud Monitoring API issue; c) quota configuration issues.
2) When MSSQL has issue but Ops Agent works as expected.

If we cannot have alert policies for scenarios 1) + 2) in the short term, in general I agree with what you proposed with having `evaluationMissingData: EVALUATION_MISSING_DATA_NO_OP` and `autoClose` as 1h. 

Otherwise, I think we can tailor the 3 alert policies proposed here to not have an open incidents when scenarios 1) and/or 2) happen. 

For scenario 1), ideally, I think there should be a standalone alert policy covers it. This alert policy alerts users when Cloud Monitoring doesn't see points flown from this Ops Agent - probably a policy with single metric-absence condition based on `agent.googleapis.com/agent/uptime`. 

For scenario 2), I also think we should have an alert policy which alerts users that there are issue collecting metrics from MSSQL. It seems that there is no existing metric descriptor representing whether MSSQL is up for Ops Agent to collect metrics yet. I don't know if there is a plan, and what the plan is. If I'd like to have an example in the OSS world, I think this kind of alert policy is similar to [this Prometheus Alert example](https://github.com/prometheus/prometheus/blob/57f4aab27d53872b109c8513846765df6b51eb57/docs/configuration/alerting_rules.md?plain=1#L63-L71).

I think the alert policy for scenario 2) can be added in this directory (though maybe not in this PR as I don't think there is a metric for scenario 2) yet).

Once a user can have an alert policy for scenario 1) + 2), the 3 proposed alert policies can keep using `evaluationMissingData=EVALUATION_MISSING_DATA_NO_OP`; or they may use `evaluationMissingData=EVALUATION_MISSING_DATA_INACTIVE`, as users can be alerted with the true root cause of scenarios 1) + 2), and investigated those failure mode specifically without needing to investigating the 3 alert policies based on the emitted metrics of MSSQL.

In conclusion, I just want to say the current approach seems good for me to be merged as it is now, if there is no plan to cover scenarios 1) + 2) in the short term. 

Once we have alert policies for scenarios 1) + 2), we can better tune the 3 alert policies here.

adeychman:
@schmikei Thank you! I think this is good to merge in 👍

EvanSimpson (approved):

PR #162, opened 2022-05-31
Add JVM alert policies

schmikei:
Proposing to add some of these JVM metrics as alerts.

Includes these alerts:

## Heap Memory Usage Near Max

Alerts on JVM heap memory usage being near maximum. When the JVM reaches its max, depending on JVM configuration could lead to Out of Memory errors.

## JVM Thread Near Limit

The JVM can spin up threads depending on the type of application. This means that this alert requires some more contextual decision-making on the threshold. When the thread count exceeds a limit (whether set by an OS or JVM configuration), then new threads cannot be spun up even if the JVM is well-provisioned.


alerts/jvm/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Heap Memory Usage Near Max
+
+Alerts on JVM heap memory usage being near maximum. When the JVM reaches its max, depending on JVM configuration could lead to Out of Memory errors.
+
+## JVM Thread Near Limit
+
+The JVM can spin up threads depending on the type of application. This means that this alert requires some more contextual decision-making on the threshold. When the thread count exceeds a limit (whether set by an OS or JVM configuration), then new threads cannot be spun up even if the JVM is well-provisioned.
+
+### Creating notification Channels and User Labels
```
adeychman, line 11: small nit: please capitalize the `notifications` in the header for consistency. Thanks!

alerts/jvm/heap-memory-usage.json
```diff
@@ -0,0 +1,26 @@
+{
+    "displayName": "JVM - Heap Memory Usage Near Max",
+    "documentation": {
+        "content": "Alerts on JVM heap memory usage being near maximum. When the JVM reaches its max, depending on JVM configuration could lead to Out of Memory errors.",
```
adeychman, line 4: Minor edit: 

"Alert for JVM heap memory usage being near maximum. When the JVM reaches its max, depending on the JVM configuration, this could lead to Out of Memory errors.",

alerts/jvm/thread-count.json
```diff
@@ -0,0 +1,36 @@
...
+        "content": "The JVM can spin up threads depending on the type of application. This means that this alert requires some more contextual decision-making on the threshold. When the thread count exceeds a limit (whether set by an OS or JVM configuration), then new threads cannot be spun up even if the JVM is well-provisioned. ",
+        "mimeType": "text/markdown"
+    },
+    "userLabels": {},
+    "conditions": [
+        {
+            "displayName": "JVM - Thread Count High",
+            "conditionThreshold": {
+                "filter": "resource.type = \"gce_instance\" AND metric.type = \"workload.googleapis.com/jvm.threads.count\"",
+                "evaluationMissingData": "EVALUATION_MISSING_DATA_NO_OP",
```
chanseokoh, line 13: Conditions setting evaluation_missing_data must have a non-zero duration.

alerts/jvm/heap-memory-usage.json
```diff
@@ -0,0 +1,26 @@
...
+    "documentation": {
+        "content": "Alert for JVM heap memory usage being near maximum. When the JVM reaches its max, depending on the JVM configuration, this could lead to Out of Memory errors.",
+        "mimeType": "text/markdown"
+    },
+    "userLabels": {},
+    "conditions": [
+        {
+            "displayName": "JVM - Heap Memory Near Max",
+            "conditionMonitoringQueryLanguage": {
+                "duration": "1m",
```
chanseokoh, line 12: condition_monitoring_query_language.duration' (type.googleapis.com/google.protobuf.Duration), Field 'duration', Illegal duration format; duration must end with 's'

However, I don't think this field must be non-zero.

schmikei: @chanseokoh do you happen to know if there is a good way to validate all the protobuf declarations of these rather than hitting an account that already has the descriptors set up? 

Just looking to save your time so I don't have to bug you to validate!

chanseokoh: I am simply copy-pasting a json into the "Try this method" tool at https://cloud.google.com/monitoring/api/ref_v3/rest/v3/projects.alertPolicies/create to call the create API against my test project. For me, I was able to make multiple calls with the same json successfully without deleting existing ones. (I just clean them later via Cloud Console.)

image

image

chanseokoh: BTW, what I noticed is that the `conditionThreshold.duration` must not be zero only when `evaluation_missing_data` is set. In this case, it's `conditionMonitoringQueryLanguage.duration` and not affected, so you can leave it to 0s if that's right.

chanseokoh (approved):

PR #168, opened 2022-06-09
Add Ops agent Nginx alert polices

JonathanWamsley:

## High connections dropped alert
Connections dropped value is derived from the connections accepted minus the connections handled. This value should be near 0. When this value is rising, then this means you may have a resource saturation problem.

## High request rate alert
The request rate is derived from the requests metrics taken as a rate every 5 minutes. This value should be monitored beforehand to understand what qualifies as a normal request rate so a threshold can be established. When request rate is above this threshold, then that means there is a large spike in traffic which logs can help diagnose if these are nefarious requests.

The `"thresholdValue"` can be adjusted based adjusted depending on what is considered to be a high request rate.

## Low request rate alert
The request rate is derived from the requests metrics taken as a rate every 5 minutes. This value should be monitored beforehand to understand what qualifies as a normal request rate so a threshold can be established. When request rate is below this threshold, then that means there is an environment problem that are limiting the request rates.

The `"thresholdValue"` can be adjusted based adjusted depending on what is considered to be a low request rate.

## High server error rate alert
The server error rate is derived from processing access logs status code. The server error rate value is the number of 5xx status codes / total status codes per 5 minutes. When the server error rate spikes suddenly, then you have a high priority server problem as clients are experiencing 5xx errors and are not being served successfully.


alerts/nginx/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+```json
+{ 
+  "userLabels": {
+    "datacenter": "central"
+  }
+}
+```
+
+## High connections dropped alert
```
adeychman, line 17: Could we move the description of the alert policies above the implementation details (notifications and user labels)? Thanks!

alerts/nginx/README.md
```diff
@@ -0,0 +1,42 @@
...
+Connections dropped value is derived from the connections accepted minus the connections handled. This value should be near 0. When this value is rising, then this means you may have a resource saturation problem.
+
+## High request rate alert
+The request rate is derived from the requests metrics taken as a rate every 5 minutes. This value should be monitored beforehand to understand what qualifies as a normal request rate so a threshold can be established. When request rate is above this threshold, then that means there is a large spike in traffic which logs can help diagnose if these are nefarious requests.
```
adeychman, line 21: missing an `of`, "as a rate of every 5 minutes"

alerts/nginx/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+The "thresholdValue" can be adjusted based adjusted depending on what is considered to be a high request rate.
+
+## Low request rate alert
+The request rate is derived from the requests metrics taken as a rate every 5 minutes. This value should be monitored beforehand to understand what qualifies as a normal request rate so a threshold can be established. When request rate is below this threshold, then that means there is an environment problem that are limiting the request rates.
```
adeychman, line 26: same minor nit, "rate of every 5 minutes"

alerts/nginx/README.md
JonathanWamsley: fixed!

alerts/nginx/README.md
JonathanWamsley: updated

alerts/nginx/README.md
JonathanWamsley: fixed!

alerts/nginx/high-server-error-rate.json
```diff
@@ -0,0 +1,37 @@
...
+        "aggregations": [
+          {
+            "alignmentPeriod": "300s",
+            "crossSeriesReducer": "REDUCE_SUM",
+            "perSeriesAligner": "ALIGN_RATE"
+          }
+        ],
+        "comparison": "COMPARISON_GT",
+        "duration": "0s",
+        "filter": "metric.type=\"logging.googleapis.com/user/nginx.server.error.count\"",
```
chanseokoh, line 21: specify a restriction on \"resource.type\" in the filter.

alerts/nginx/low-request-rate.json
```diff
@@ -0,0 +1,37 @@
...
+          "thresholdValue": 10,
+          "trigger": {
+            "count": 1
+          }
+        }
+      }
+    ],
+    "alertStrategy": {
+      "autoClose": "604800s",
+      "notificationPrompts": [
```
chanseokoh, line 30: Property "notificationPrompts" does not exist in object specification.

alerts/nginx/high-server-error-rate.json
JonathanWamsley: updated

alerts/nginx/low-request-rate.json
JonathanWamsley: updated

chanseokoh (approved):

PR #175, opened 2022-06-16
Add Ops agent Hadoop alert polices

JonathanWamsley:

## Low available capacity alert
A low available capacity means that the total available disk space capacity across all HDFS clusters is low and should be increased.

## Volume failure alert
A failed volume means a hardware failure has occurred. With replication, no data should be lost, but the failed hardware should be replaced.

## Dead data nodes alert
A dead data node can cause network activity problems as the NameNode initiates replication of blocks lost on the data nodes.When multiple data nodes are lost, data loss can occur.

### DIFF
--- a/alerts/hadoop/README.md
+++ b/alerts/hadoop/README.md
@@ -1,5 +1,14 @@
 # Alerts for Hadoop in the Ops Agent
 
+## Low available capacity alert
+A low available capacity means that the total disk space usage across all HDFS clusters is over 80% of the capacity limit and therefore the capacity limit should be increased.
+
+## Volume failure alert
+A failed volume means a hardware failure has occurred. With replication, no data should be lost, but the failed hardware should be replaced.
+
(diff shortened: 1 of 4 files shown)


alerts/hadoop/README.md
```diff
@@ -0,0 +1,24 @@
...
+```json
+{ 
+  "userLabels": {
+    "datacenter": "central"
+  }
+}
+```
+
+## Low available capacity alert
+A low available capacity means that the total available disk space capacity across all HDFS clusters is low and should be increased.
```
adeychman, line 18: These descriptions are great, thank you. Could you also include what threshold is being used in the alert policy sample? i.e. how is "low" being defined?

JonathanWamsley: added in threshold details and organized the section

chanseokoh (approved):

PR #177, opened 2022-06-17
Add Ops agent RabbitMQ alert polices 

JonathanWamsley:

## High unacknowledged messages alert
When the message unacknowledged mean is higher than a threshold, then errors or performance issues may arise.

## High unroutable messages alert
When the unroutable message rate is greater than a threshold, then the consumers are not getting every message.

## Low deliverable messages alert
When the deliverable message rate are lower than a threshold, then there may be an issue with a producer or routing logic.

### DIFF
--- a/alerts/rabbitmq/README.md
+++ b/alerts/rabbitmq/README.md
@@ -1,0 +1,24 @@
+# RabbitMQ Alerts for Ops Agent
+
+### Notification Channels
+For all alerts, a notification channel needs to be set up or the alert will fire silently.
+
+### User Labels
+User labels can be used for these policies by modifying the userLabels fields of the policies. i.e.
+
+```json
+{ 
+  "userLabels": {
+    "datacenter": "central"
+  }
+}
+```
+
+## High unacknowledged messages alert
+When the message unacknowledged mean is higher than a threshold, then errors or performance issues may arise.
+
+## High unroutable messages alert
+When the unroutable message rate is greater than a threshold, then the consumers are not getting every message.
+
+## Low deliverable messages alert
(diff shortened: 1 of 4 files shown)


alerts/rabbitmq/README.md
```diff
@@ -0,0 +1,24 @@
...
+```
+
+## High unacknowledged messages alert
+When the message unacknowledged mean is higher than a threshold, then errors or performance issues may arise.
+
+## High unroutable messages alert
+When the unroutable message rate is greater than a threshold, then the consumers are not getting every message.
+
+## Low deliverable messages alert
+When the deliverable message rate are lower than a threshold, then there may be an issue with a producer or routing logic.
```
adeychman, line 24: "are lower" --> "is lower"

alerts/rabbitmq/rabbitmq-low-deliverable-messages alert.json
```diff
@@ -0,0 +1,34 @@
+{
+  "displayName": "Rabbitmq - low deliverable messages alert",
+  "documentation": {
+    "content": "When the deliverable message rate are lower than a threshold, then there may be an issue with a producer or routing logic.",
```
adeychman, line 4: "are lower" --> "is lower"

alerts/rabbitmq/README.md
JonathanWamsley: fixed!

alerts/rabbitmq/rabbitmq-low-deliverable-messages alert.json
JonathanWamsley: fixed!

chanseokoh (approved):

PR #180, opened 2022-06-21
Add Ops agent Tomcat alert polices

JonathanWamsley:

## High request rate alert
The request rate is derived from the requests metrics taken as a rate every 5 minutes. This value should be monitored beforehand to understand what qualifies as a normal request rate so a threshold can be established. When request rate is above this threshold, then that means there is a large spike in traffic which logs can help diagnose if these are nefarious requests.

## Low request rate alert
The request rate is derived from the requests metrics taken as a rate every 5 minutes. This value should be monitored beforehand to understand what qualifies as a normal request rate so a threshold can be established. When request rate is below this threshold, then that means there is an environment problem that are limiting the request rates.

## High server error rate alert
The server error rate is derived from processing access logs status code. The server error rate value is the number of 5xx status codes per 5 minute window. When the server error rate spikes suddenly, then you have a high priority server problem as backends are experiencing 5xx errors and clients are not being served successfully.

### DIFF
--- a/alerts/tomcat/README.md
+++ b/alerts/tomcat/README.md
@@ -1,19 +1,5 @@
 # Alerts for Tomcat in the Ops Agent
 
-### Notification Channels
-For all alerts, a notification channel needs to be set up or the alert will fire silently.
-
-### User Labels
-User labels can be used for these policies by modifying the userLabels fields of the policies. i.e.
-
-```json
-{ 
-  "userLabels": {
-    "datacenter": "central"
-  }
-}
-```
-
 ## High request rate alert
 The request rate is derived from the requests metrics taken as a rate every 5 minutes. This value should be monitored beforehand to understand what qualifies as a normal request rate so a threshold can be established. When request rate is above this threshold, then that means there is a large spike in traffic which logs can help diagnose if these are nefarious requests.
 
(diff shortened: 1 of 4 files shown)


alerts/tomcat/README.md
```diff
@@ -0,0 +1,39 @@
...
+  "userLabels": {
+    "datacenter": "central"
+  }
+}
+```
+
+## High request rate alert
+The request rate is derived from the requests metrics taken as a rate every 5 minutes. This value should be monitored beforehand to understand what qualifies as a normal request rate so a threshold can be established. When request rate is above this threshold, then that means there is a large spike in traffic which logs can help diagnose if these are nefarious requests.
+
+The `"thresholdValue"` can be adjusted based adjusted depending on what is considered to be a high request rate.
```
adeychman, line 20: typo, double reference to adjusted.

JonathanWamsley: fixed!

alerts/tomcat/tomcat-high-server-error-rate.json
```diff
@@ -0,0 +1,35 @@
...
+        "aggregations": [
+          {
+            "alignmentPeriod": "300s",
+            "crossSeriesReducer": "REDUCE_SUM",
+            "perSeriesAligner": "ALIGN_RATE"
+          }
+        ],
+        "comparison": "COMPARISON_GT",
+        "duration": "0s",
+        "filter": "metric.type=\"logging.googleapis.com/user/tomcat.server.error.count\"",
```
chanseokoh, line 21: resource.type must be specified.

JonathanWamsley: updated

chanseokoh (approved):

PR #183, opened 2022-06-21
Add Memcached Alerts

Mrod1598:
Alerts to be added:
## High CPU Usage

if `memcached.cpu.usage` is higher than what is anticipated it could show that the Server us overloaded and efforts to improve performance may be necessary. Subjective to environment, configure threshold knowingly.

## High Evictions

if `memcached.evictions` increase above `0` that shows that items are being evicted due to high memory usage and should be a cause for concern.

## No Connections

if `memcached.connections.current` reaches `0` that should be cause for concern because memcached should always have at least 1 connection.

### DIFF
--- a/alerts/memcached/README.md
+++ b/alerts/memcached/README.md
@@ -31,7 +31,7 @@ i.e.
 
 #### Notification Channels
 
-The ID of the notification channel to be notifed.
+The ID of the notification channel to be notified.
 
 ```json
     "notificationChannels": [
--- a/alerts/memcached/README.md
+++ b/alerts/memcached/README.md
@@ -1,0 +1,40 @@
+
+# Alerts for Memcached in the Ops Agent
+
+## High CPU Usage
+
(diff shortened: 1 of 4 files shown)


alerts/memcached/README.md
```diff
@@ -0,0 +1,40 @@
+
+# Alerts for Memcached in the Ops Agent
+
+## High CPU Usage
+
+if `memcached.cpu.usage` is higher than what is anticipated it could show that the Server us overloaded and efforts to improve performance may be necessary. Subjective to environment, configure threshold knowingly.
```
adeychman, line 6: Please capitalize the first letter in the descriptions

adeychman: typo, `us` --> `is`

adeychman: Could you please also include in the description what threshold is being used in the sample alert policy?

Mrod1598: Fixed

chanseokoh (approved):

PR #185, opened 2022-06-22
Add Solr Alerts

Mrod1598:
Alerts to be added: 

## High Request Errors

If `solr.request.error.count` is high there should be cause for concern as your server could be having issue executing queries. 

## High Cache Evictions

If `solr.cache.eviction.count` is high, Solr is having trouble keeping up. Either the cache eviction time should be increased or improvements in performance should be looked for.

## High Request Count

If `solr.request.count` is above a threshold that should be determined based on your server, it could show there is a security issue.

### DIFF
--- a/alerts/solr/README.md
+++ b/alerts/solr/README.md
@@ -6,11 +6,11 @@ If `solr.request.error.count` is high there should be cause for concern as your 
 
 ## High Cache Evictions
 
-If `solr.cache.eviction.count` is high, Solr is having trouble keeping up. Either the cache eviction time should be increased or improvements in performance should be looked for.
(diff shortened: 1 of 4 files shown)


alerts/solr/README.md
```diff
@@ -0,0 +1,39 @@
+# Alerts for Solr in the Ops Agent
+
+## High Request Errors
+
+If `solr.request.error.count` is high there should be cause for concern as your server could be having issue executing queries.
```
adeychman, line 5: In the descriptions for the ReadMe and in the JSON, could you share what thresholds are being set in the sample alert policy? 

i.e. how is "high" being defined?

Mrod1598: Updated

chanseokoh (approved):

PR #188, opened 2022-06-22
Add Zookeeper Alerts

Mrod1598:
Alerts to be added:

## High Average Latency

if `zookeeper.latency.avg` is 100ms or higher, it shows that the server can't keep up with demand and performance improvements should be sought after.

## High fsync Duration

if `zookeeper.fsync.exceeded_threshold.count` is above an amount dependent on your environment it could show either the fsync duration  threshold set is too low or there is an issue writing to files.

## Low Open File Descriptors

if `workload.googleapis.com/zookeeper.file_descriptor.open` is about to hit `workload.googleapis.com/zookeeper.file_descriptor.limit` zookeeper will not be able to handle any new requests and will fail to connect new clients.

### DIFF
--- a/alerts/zookeeper/README.md
+++ b/alerts/zookeeper/README.md
@@ -2,15 +2,15 @@
 
 ## High Average Latency
 
-if `zookeeper.latency.avg` is 100ms or higher, it shows that the server can't keep up with demand and performance improvements should be sought after.
+If `zookeeper.latency.avg` is 100ms or higher, it shows that the server can't keep up with demand and performance improvements should be sought after.
 
 ## High fsync Duration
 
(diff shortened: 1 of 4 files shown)


alerts/zookeeper/README.md
```diff
@@ -0,0 +1,39 @@
+# Alerts for Zookeeper in the Ops Agent
+
+## High Average Latency
+
+if `zookeeper.latency.avg` is 100ms or higher, it shows that the server can't keep up with demand and performance improvements should be sought after.
```
adeychman, line 5: Small nit: Please capitalize the first letter of the sentence in this description and below

Mrod1598: Updated

chanseokoh (approved):

PR #189, opened 2022-06-22
Add WildFly Alert Policies

schmikei:
Proposing to add these alerts:

# Alert Policies for Wildfly in the Ops Agent

## Server Errors

If 5 or more server errors are happening over a 1 minute interval could be an indication that something is programatically incorrect or could use further exploration into why the server errors are occurring.

## Waiting Requests

Waiting requests could be an indication of high CPU usage or perhaps slow disk I/O. It would be recommended to use some kind of CPU utilization tool like top to dig deeper.

## JDBC Connections near limit

If the JDBC connection limit is reaching its max (default=20 connections), then backend connections could end up failing.


alerts/wildfly/README.md
```diff
@@ -0,0 +1,39 @@
...
+    }
+```
+
+#### Notification Channels
+
+The ID of the notification channel to be notified.
+
+```json
+    "notificationChannels": [
+        "projects/project-id/notificationChannels/1234567
```
antonblock, line 37:
```suggestion
        "projects/project-id/notificationChannels/1234567"
```

alerts/wildfly/README.md
```diff
@@ -0,0 +1,39 @@
+# Alert Policies for Wildfly in the Ops Agent
+
+## Server Errors
+
+If 5 or more server errors are happening over a 1 minute interval could be an indication that something is programatically incorrect or could use further exploration into why the server errors are occurring.
+
+## Waiting Requests
+
+Waiting requests could be an indication of high CPU usage or perhaps slow disk I/O. It would be recommended to use some kind of CPU utilization tool like top to dig deeper.
```
adeychman, line 9: Could we indicate the threshold that's used in the alert policy? For example, "If X or more requests are queued for more than Y minutes, then the alert policy is triggered"

alerts/wildfly/server-errors.json
```diff
@@ -0,0 +1,36 @@
...
+    "displayName": "WildFly - Server Errors",
+    "documentation": {
+      "content": "Raw server errors coming form WildFly could indicate application outages/other uptime factors.",
+      "mimeType": "text/markdown"
+    },
+    "userLabels": {},
+    "conditions": [
+      {
+        "displayName": "Wildfly - Server Errors",
+        "evaluationMissingData": "EVALUATION_MISSING_DATA_NO_OP",
```
chanseokoh, line 11: Property "evaluationMissingData" does not exist in object specification.

alerts/wildfly/server-errors.json
```diff
@@ -0,0 +1,36 @@
...
+          "filter": "resource.type = \"gce_instance\" AND metric.type = \"workload.googleapis.com/wildfly.request.server_error\"",
+          "aggregations": [
+            {
+              "alignmentPeriod": "60s",
+              "crossSeriesReducer": "REDUCE_NONE",
+              "perSeriesAligner": "ALIGN_RATE"
+            }
+          ],
+          "comparison": "COMPARISON_GT",
+          "duration": "10s",
```
chanseokoh, line 22: only multiples of 1 minute are supported

PR #193, opened 2022-06-23
Add Ops agent MySQL alert polices

JonathanWamsley:

## High number of log warnings alert
Log warnings are a count of logs collected from mysql_error logs with a warning level per 5 minute window. The default condition of warnings per 5 minute window is 20.

### Prerequisites
The name of the metric should be:
`mysql.warning.count`

The filter for the metric should be:
```
logName:"mysql_error"
jsonPayload.level="Warning"
```

## High number of log errors alert
Log error are a count of logs collected from mysql_error logs with an error level per 5 minute window. The default condition of error per 5 minute window is 5.

### Prerequisites
The name of the metric should be:
`mysql.error.count`

The filter for the metric should be:
```
logName:"mysql_error"
jsonPayload.level="Error"
```

## High slow queries rate
The slow queries are collected from mysql_slow logs. More specific queries can be filtered using the `jasonPayload.message` field shown in the filter below. Additionally, different max query times can also be filtered using `jsonPayload.queryTime` field.

### Prerequisites
The name of the metric should be:
`mysql.slow_queries.count`

The filter for the metric should be:
```
logName:"mysql_slow"
jsonPayload.message:("SELECT" OR "INSERT" OR "UPDATE" OR "CREATE" OR "DELETE")
jsonPayload.queryTime>1
```

### DIFF
--- a/alerts/mysql/mysql-high-slow-queries-rate.json
+++ b/alerts/mysql/mysql-high-slow-queries-rate.json
@@ -17,7 +17,7 @@
         ],
         "comparison": "COMPARISON_GT",
         "duration": "0s",
-        "filter": "metric.type=\"logging.googleapis.com/user/mysql.slow_queries.count\"",
+        "filter": "resource.type = \"gce_instance\" AND metric.type=\"logging.googleapis.com/user/mysql.slow_queries.count\"",
         "thresholdValue": 0.5,
         "trigger": {
           "count": 1
--- a/alerts/mysql/mysql-high-slow-queries-rate.json
+++ b/alerts/mysql/mysql-high-slow-queries-rate.json
@@ -1,0 +1,34 @@
+{
+  "displayName": "MySQL - high slow queries rate",
+  "documentation": {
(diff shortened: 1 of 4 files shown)


alerts/mysql/mysql-high-slow-queries-rate.json
```diff
@@ -0,0 +1,34 @@
...
+      "conditionThreshold": {
+        "aggregations": [
+          {
+            "alignmentPeriod": "300s",
+            "perSeriesAligner": "ALIGN_RATE"
+          }
+        ],
+        "comparison": "COMPARISON_GT",
+        "duration": "0s",
+        "filter": "metric.type=\"logging.googleapis.com/user/mysql.slow_queries.count\"",
```
chanseokoh, line 20: resource.type must be specified.

JonathanWamsley: updated

chanseokoh (approved):

PR #215, opened 2022-07-01
LogsPanel dev docs guidance and NGINX dash update

draffensperger:
@EvanSimpson and @xiangshen-dk The Monitoring UI LogsPanel now auto-infers the project ID for the logs based on the currently selected metrics scope and so this is a good time to update the guidance about using LogsPanel in sample dashboards. I've also updated the NGINX dashboard as an example of this.


dev-docs/DASHBOARD_TIPS.md
```diff
@@ -6,29 +6,25 @@ docs.
...
-project, which would be misleading to the customer as they wonder where their
-e.g. NGINX logs went, when in fact they are there just not being shown.
-
-As of 2021-12-14 this is the case, but we will investigate how we could
-modify the LogsPanel component to enable it to be more flexible and have it e.g.
-infer the `resourceNames` from the current metric scope if it's unspecified.
-
-In the mean time, the best we can do is to link to logs in a markdown card,
-see below under "Embedding Markdown With Links to Docs/Logs".
+In order to do that, use the `log_id` function to filter for the suffix part of
```
EvanSimpson, line 21: As of Ops Agent 2.17.0, integration logs have a ground-truth label that can be queried instead of the user-configurable log name.

An example filter would instead look like
```
labels."logging.googleapis.com/instrumentation_source"="agent.googleapis.com/nginx_access" resource.type="gce_instance
```
where the `nginx_access` part is the log type, and is not user-configurable.

Can you update the examples here to use that instead, and also add a markdown widget to the dashboard mentioning the minimum version requirement for the logs to work (which for some integrations will be higher than the minimum version to collect logs without this label).

EvanSimpson (changes requested):

dev-docs/DASHBOARD_TIPS.md
draffensperger: Yeah, that makes sense. What if instead of specifying a minimum version for the logs to be supported, say we instead do an `OR` with the default log ID? That could be a way to sort of be best-effort backwards compatible and then might not warrant a disclaimer about the agent version. I've updated the PR with that idea. Happy to do with just the label only if you prefer though. WDYT?

EvanSimpson: Yeah, I think that works! It should only be necessary for integrations introduced before 2.17.0, but it's a nice fallback.

PR #270, opened 2022-08-19
Add filtered dashboards for GKE per-cluster, node, workload views

yqlu:


dashboards/compute/gke-compute-resources-workload-view.json
```diff
@@ -0,0 +1,782 @@
...
+            }
+          }
+        },
+        "width": 6,
+        "xPos": 6,
+        "yPos": 22
+      }
+    ]
+  },
+  "name": "projects/180739612668/dashboards/693f745f-fab1-40c1-9f99-c97507e1c10c"
```
EvanSimpson, line 781: Can you remove this line?

EvanSimpson (changes requested):

dashboards/compute/gke-compute-resources-workload-view.json
yqlu: Good catch, done

EvanSimpson (approved):

PR #271, opened 2022-08-23
Split up dashboards/compute to 4 folders

yqlu:


dashboards/google-cloud-run/metadata.yaml
```diff
@@ -0,0 +1,31 @@
+sample_dashboards:
+  -
+    category: Compute
```
EvanSimpson, line 3: nit: Cloud Run for the categories in this file.

EvanSimpson (commented):
@xiangshen-dk FYI - this change splits the "compute" dashboards into separate folders for each product.

dashboards/google-cloud-run/metadata.yaml
yqlu: Good catch, fixed (and deleted the unnecessary entries in this file)

EvanSimpson (approved):

PR #272, opened 2022-08-23
Initial GKE alerting policies

BinaryFissionGames:
This PR contains the initial GKE alerting policies based on prometheus metrics.
Most of these are just ports of the ops agent alerting policies, with the exception of:
* `Kafka - Prometheus - Change in Number of ISRs`
* `MySQL - Prometheus - Connection Errors`

This PR contains alerting policies for:
* Elasticsearch
* Kafka
* Memcached
* MongoDB
* MySQL
* NGINX
* PostgreSQL
* RabbitMQ
* Redis
* Zookeeper

Exact location of these polices are still to be determined.


alerts/elasticsearch-gke/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Red cluster status alert
+When the cluster has a "red" status, at least one primary shard is missing, and data is missing.
+
+## High JVM memory heap usage alert
+When the ratio of heap used to max heap becomes high, then application's performance may start to degrade.
+
+### Creating notification Channels and User Labels
+
+Whether these alert policies are being used as standalones or base templates for a deployment strategy like terraform, one thing that should be utilized is notification channels and user labels.
```
adeychman, line 14: Let's update this to explain why it should be used: 

"It is strongly recommended to add notification channels and user labels to the alert policies. The notification channel will set the notification destination if the alert policy is triggered. User labels are used for categorization, and can be used to indicate the severity level." 

Could you please make this edit in the readme's below as well?

alerts/elasticsearch-gke/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+### User Labels
+
+Supplying user labels could give extra identification information about the firing alert:
+
+i.e.
+
+```json
+    "userLabels": {
+        "datacenter": "central"
```
adeychman, line 24: Could we update this to be severity level? i.e. "Severity" : "Warning".

Since the datacenter/location is typically automatically available from the underlying monitored resource. 

Could you please make this edit in the readme's below as well?

alerts/kafka-gke/README.md
```diff
@@ -0,0 +1,33 @@
+# Kafka Alerts for GKE
+
+## Under replicated partitions alert
+When there is an under replicated partition for an extended period of time, then the cluster is not healthy and cannot guarantee high availability.
+
+## Change in number of ISRs alert
+In sync replicas (ISRs) are required for proper failover. The number of in sync replicas should be static, unless the broker cluster is expanding or partitions are removed. If the number of in sync replicas deviates without the broker cluster being modified, you should investigate to maintain high availability.
```
adeychman, line 7: Is  "In Sync Replicas" a term? If so, how about we capitalize it here and in the descriptions below?

alerts/postgres-gke/README.md
```diff
@@ -0,0 +1,31 @@
+# Alerts for PostgreSQL in GKE
+
+## Database size too large alert
+If the database is growing larger than expected (100 GB by default), it may signal excess data being stored or excess records not being properly removed by an application.
```
adeychman, line 4: Is it better to be a static value or a percentage? I'm wondering if the static value covers majority of use cases

alerts/postgres-gke/README.md
```diff
@@ -0,0 +1,31 @@
+# Alerts for PostgreSQL in GKE
```
adeychman, line 1: Is there 1 recommended alert policy for Postgres?

alerts/elasticsearch-gke/README.md
```diff
@@ -0,0 +1,36 @@
+# Elasticsearch Alerts for GKE
```
adeychman, line 1: As a general question, what are your thoughts about adding potential resolution steps in the documentation? Are there any common best practices that if the failure occurs, what are common remediation steps?

adeychman (commented):
Thank you! I left some comments, PTAL

ddli-google:
Hi Team!

In order for alerts to appear in the Cloud Console, we are introducing a set of guidelines on changes to the /alerts folder.  These are largely reminiscent of those for integrations & dashboards. For this PR and moving forward, we would like to implement the following requirements:

  1. Each folder should have an accompanying metadata.yaml file describing each json file in the same folder.
  2. All alert policy json file names are in the format <name>.v<version_number>.json (ex: foo.v1.json).
  3. When making revisions to a policy, do not directly modify a json file. Instead, upload as a new file of the same name with 
      the version number increased by one (ex: foo.v1.json -> foo.v2.json).

You can find examples at https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/tree/master/alerts/google-cloud-alerting.

ObservIQ does not need to worry about the existing folders/files; we will backfill the metadata.yaml and modify the json file names for those.

alerts/postgres-gke/README.md
BinaryFissionGames: A percentage would be preferable, but there doesn't seem to be a good way to get the limit (max # of bytes available to postgres) to compute the percentage, so we've kept it a static limit.

alerts/elasticsearch-gke/README.md
BinaryFissionGames: Generally, remediation steps fall outside of this scope of work. More generally, we're more hesitant to recommend 'common' remediation steps due to the risk/variability of customer environments.

alerts/postgres-gke/README.md
BinaryFissionGames: Yes; We have 3 planned alert policies for ops-agent, but 2 of them are derived from metrics not included in the prometheus exporter, unfortunately, so they have been dropped here.

BinaryFissionGames:
@ddli-google Thanks for the info! I've added in the metadata.yaml and changed the file names to conform with the versioning.

@adeychman I think I've addressed all the feedback you've had in your initial pass, I think it's ready for another look when you get the chance.

chanseokoh:
I've done syntax reviews on all the `.json` files. They look good to me.

@adeychman I'll leave the approval up to you.

adeychman:
@BinaryFissionGames LGTM! Thank you

EvanSimpson (approved):

PR #280, opened 2022-08-31
fix: MSSQL alerts: Remove evaluationMissingData field

BinaryFissionGames:
The presence of this field causes the policy to fail to import with the following error:
```
Field alert_policy.conditions[0].condition_threshold.evaluation_missing_data had an invalid value of "3": Conditions setting evaluation_missing_data must have a non-zero duration.
```

If `duration` is `0s`, this field is not allowed.

### DIFF
--- a/alerts/mssql-server/connections-near-limit.v1.json
+++ b/alerts/mssql-server/connections-near-limit.v1.json
@@ -19,7 +19,6 @@
                 ],
                 "comparison": "COMPARISON_GT",
                 "duration": "0s",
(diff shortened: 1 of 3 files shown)


cocosheng (approved):

EvanSimpson:
@BinaryFissionGames do you mind resolving the file name conflicts here?

BinaryFissionGames:
@EvanSimpson Rebased; Let me know if we actually want these in a v2. I think it's better here to just fix v1, since it was an invalid alert (couldn't be imported).

EvanSimpson:
I think keeping this as v1 is fine, for the reason you point out.

PR #281, opened 2022-08-31
Ops Agent Alerts for PostgreSQL

schmikei:
Alerts proposed:

## High CPU Utilization

Alerts whenever the CPU utilization goes above 80% which usually indicates the instance's performance is heavily degraded and likely is going to impact applications reliant on postgres.

## Connections Near Limit

Connections are a limiting factor in postgres and once the instance cannot accept any more connections, then applications reliant on the postgres instance cannot function.

## High Database Size

Alert fires when the database size is growing greater than expected (this value will be subject to instance size and utilization); defaulted to 93 GB but will be subject to instance size as well as connected storage.

## Reached Max Written Buffers

Alert fires when the background writer attains 3 maxwritten errors i.e. it cannot flush buffers because it has written too many. This is an indication that if an outage were to occur, then any bytes pending may be subject to be lost.

Submitting in draft until metric additions get merged :)

### DIFF
--- a/alerts/postgresql/README.md
+++ b/alerts/postgresql/README.md
@@ -1,0 +1,17 @@
+# Alerts for Postgres in Ops Agent
+
+## High CPU Utilization
+
+Alerts whenever the CPU utilization goes above 80% which usually indicates the instance's performance is heavily degraded and likely is going to impact applications reliant on postgres.
+
+## Connections Near Limit
+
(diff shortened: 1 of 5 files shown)


schmikei:
With https://github.com/GoogleCloudPlatform/ops-agent/pull/804 merged, it is probably safe to merge this in the next release

adeychman:
LGTM

EvanSimpson (approved):

PR #302, opened 2022-09-14
Initialize quotas alert policy folder with exceeded template

rheewalt:
Initializing the quotas alert policy folder and added a alert policy template for exceeded quota alerts.

Screen Shot 2022-09-14 at 6 05 25 PM

Validated the template with the placeholders replaced with valid values.

### DIFF
--- a/alerts/google-quotas/exceeded-quota.v1.json
+++ b/alerts/google-quotas/exceeded-quota.v1.json
@@ -1,0 +1,27 @@
+{
+  "displayName": "Quota Exceeded Alert Policy",
+  "userLabels": {},
+  "conditions": [
+    {
+      "displayName": "Quota Exceeded",
+      "conditionThreshold": {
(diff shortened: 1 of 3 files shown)


alerts/google-quotas/exceeded-quota.v1.json
```diff
@@ -0,0 +1,27 @@
+{
+  "displayName": "Quota Exceeded Alert Policy",
```
cocosheng, line 2: we suggest this value to match the display name in metadata so users won't get confused when seeing the list of created alert policies.

rheewalt: Done.

cocosheng (approved):

rheewalt:
@cocosheng Thanks for the quick review.  Can you merge this PR or assign the correct people with write access to do so?

PR #310, opened 2022-09-23
Add metadata file for cloudsql dashboards

EvanSimpson:


dashboards/google-cloudsql/metadata.yaml
```diff
@@ -0,0 +1,28 @@
+sample_dashboards:
+  -
+    category: CloudSQL
+    id: 1p-cloudsql-general
```
turnerhayes-google, line 4: I think we use gcp- as the prefix

dashboards/google-cloudsql/metadata.yaml
```diff
@@ -0,0 +1,28 @@
...
+      This dashboard has charts displaying logs from CloudSQL as well as charts of instance states and network throughput.
+    related_integrations:
+      - id: cloudsql
+        platform: GCP
+  -
+    category: CloudSQL
+    id: 1p-cloudsql-replication
+    display_name: CloudSQL Replication
+    description: |-
+      This dashboard has charts displaying information about what replication states and replication performance from CloudSQL.
```
turnerhayes-google, line 16: information about ~~what~~ replication states?

The sentence doesn't really make sense as is

turnerhayes-google (changes requested):

dashboards/google-cloudsql/metadata.yaml
EvanSimpson: Good spot! Fixed.

dashboards/google-cloudsql/metadata.yaml
EvanSimpson: This is the dashboard id which doesn't have a set convention. I just removed the prefix since I don't think it's necessary. I have also updated the integration id value to match the internal value.

turnerhayes-google (approved):

PR #313, opened 2022-09-28
Fix dashboard review rotation config

EvanSimpson:


varun-c (approved):
Will everyone be cc'ed in the mail?

EvanSimpson:
> Will everyone be cc'ed in the mail?

No, it will be up to each individual's email preferences within GitHub.

PR #336, opened 2022-10-13
Add dashboard validation workflow

philshin-google:
… yaml file


scripts/dashboard/validate_dashboards_format.py
```diff
@@ -0,0 +1,84 @@
...
+
+  except:
+    raise Exception("{} content does not match expected sample dashboard format".format(path))
+
+
+def main():
+  path = sys.argv[1]
+  # only run validation script on files added/changed in
+  # dashboards folder
+  if os.path.dirname(os.path.dirname(path)) != "dashboards":
```
mr-arc, line 64: This will behave differently whether there is a trailing slash in `path`. Is the goal to strip off the filename and go up one level? Maybe add a comment to clarify the intent here, too?

scripts/dashboard/validate_dashboards_format.py
```diff
@@ -0,0 +1,84 @@
+import json
```
mr-arc, line 1: Add a file docstring that explains how this script is to be used, including the command line arguments.

scripts/dashboard/validate_dashboards_format.py
```diff
@@ -0,0 +1,84 @@
...
+
+  file_name = os.path.basename(path)
+  file_name_parts = file_name.split(".")
+  # metadata file added/changed would be checked for expected fields
+  if file_name == "metadata.yaml":
+    check_metadata_entries(path)
+
+  # all json files added to dashboards folder must follow expected
+  # file hierarchy and naming
+  elif path.split(".")[-1] == "json":
```
mr-arc, line 75: You've already split out the filename, so you can do:

```
elif file_name_parts[-1] == "json":
```

philshin-google: done.

scripts/dashboard/validate_dashboards_format.py
philshin-google: done.

scripts/dashboard/validate_dashboards_format.py
philshin-google: refactored this line into the new mapPaths function. New equivalent line is `if not 'dashboards' == pathParts[1]:` in mapPaths(). Should avoid error / confusion detailed here.

scripts/dashboard/validate_dashboards_format.py
```diff
@@ -0,0 +1,84 @@
...
+        if "widget" in tile.keys():
+            if not "title" in tile["widget"]:
+                raise Exception("{} content does not match expected sample dashboard format".format(path))
+
+  except:
+    raise Exception("{} content does not match expected sample dashboard format".format(path))
+
+
+def main():
+  path = sys.argv[1]
```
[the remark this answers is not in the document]
philshin-google: based off the github workflow outlined in the validate_dashboards_format.yml, this script should only be triggered when there are new/added files, so there should never be an instance where there are no arguments.

scripts/dashboard/validate_dashboards_format.py
```diff
@@ -0,0 +1,84 @@
+import json
+import os
+import sys
+import yaml
+from google.cloud import monitoring_v3
+
+def check_json_in_metadata(path, file_id):
```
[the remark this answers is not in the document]
philshin-google: done.

scripts/dashboard/validate_dashboards_format.py
```diff
@@ -0,0 +1,84 @@
...
+  if len(file_name_parts) != 2:
+    raise Exception("{} file name not in <name>.json format".format(path))
+  file_version = file_name_parts[1]
+
+def check_sample_dashboards_json(path):
+  with open(path) as f:
+    try:
+      dashboards_dict = json.load(f)
+      dashboards_json = json.dumps(dashboards_dict)
+    except:
```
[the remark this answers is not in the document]
philshin-google: done.

scripts/dashboard/validate_dashboards_format.py
```diff
@@ -0,0 +1,84 @@
...
+    if tmpl.get("id") == file_id:
+      return
+  raise Exception("{} does not have an entry in {}".format(path, metadata_path))
+
+def check_metadata_entries(path):
+  with open(path) as f:
+    data = yaml.safe_load(f)
+  sample_dashboards = data.get("sample_dashboards")
+  if not sample_dashboards:
+    raise Exception("sample_dashboards not defined in {}".format(path))
```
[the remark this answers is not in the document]
philshin-google: done.

scripts/dashboard/validate_dashboards_format.py
```diff
@@ -0,0 +1,84 @@
...
+
+def check_json_in_metadata(path, file_id):
+  metadata_path = os.path.join(os.path.dirname(path), "metadata.yaml")
+  check_metadata_entries(metadata_path)
+  with open(metadata_path) as f:
+    data = yaml.safe_load(f)
+  for tmpl in data.get("sample_dashboards"):
+    if tmpl.get("id") == file_id:
+      return
+  raise Exception("{} does not have an entry in {}".format(path, metadata_path))
```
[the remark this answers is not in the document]
philshin-google: done.

scripts/dashboard/validate_dashboards_format.py
```diff
@@ -0,0 +1,153 @@
...
+
+    pathMap[directory]['jsonNames'][dashDict['displayName']]=''
+
+    fileNameParts = jsonFile.split('.')
+    check_json_file_name(jsonFile, fileNameParts)
+
+    pathMap[directory]['jsonIds'][fileNameParts[0]]=''
+
+def check_metadata_entries(path, sample_dashboard):
+  '''Check if metadata file fields are as expected'''
```
EvanSimpson, line 85: nit: Assert sample dashboard metadata entry has all required fields.

scripts/dashboard/validate_dashboards_format.py
```diff
@@ -0,0 +1,153 @@
...
+  required_fields = {"category", "id", "display_name", "description"}
+  missing_fields = required_fields - sample_dashboard.keys()
+  if missing_fields:
+    raise MetadataFormattingError("{} missing {}".format(path, missing_fields))
+
+def check_metadata(directory, pathMap):
+  '''Checks metadata file in directory for correct fields and matching data in 
+  json files within the same directory'''
```
EvanSimpson, line 93: This sentence is difficult to parse. I think this might read better as a list, e.g.

```
Make following assertions about metadata.yam files:
* At least 1 sample dashboard exists
* Each dashboard entry has all required fields
* Each dashboard entry has a corresponding json template with same name & id
```

scripts/dashboard/validate_dashboards_format.py
```diff
@@ -0,0 +1,153 @@
...
+    pathMap[directory]['metadataNames'][dashboardName]=''
+
+    if not dashboardId in pathMap[directory]['jsonIds'].keys():
+      raise MetadataInJsonMisMatchError("{} does not have a matching json file".format(dashboardId))
+    
+    if not dashboardName in pathMap[directory]['jsonNames'].keys():
+      raise MetadataInJsonMisMatchError("{} does not have a matching json file".format(dashboardName))
+
+def check_json_in_metadata(directory, pathMap):
+  '''Checks json file data against metadata file of the same directory'''
```
EvanSimpson, line 118: Similar to my comment above, this could be more specific

```
Make following assertions about json template files:
* Template file name appears in metadata.yaml file
* Template display name matches corresponding id in metadata.yaml file
```

scripts/dashboard/validate_dashboards_format.py
```diff
@@ -0,0 +1,153 @@
...
+  '''Parent function that calls top level processing and checking functions'''
+
+  # Process json files into path map
+  mapJsonFiles(directory, pathMap)
+
+  # Metadata file check
+  check_metadata(directory, pathMap)
+
+  # json file check
+  check_json_in_metadata(directory, pathMap)
```
EvanSimpson, line 139: It would be worth adding to these comments to call out that they are order dependent, since they are mutating `pathMap`.

scripts/dashboard/validate_dashboards_format.py
EvanSimpson: Would also be good to add a line that it mutates pathMap

scripts/dashboard/validate_dashboards_format.py
```diff
@@ -0,0 +1,153 @@
...
+
+def mapPaths(paths):
+  '''Takes in list of paths and returns dictionary of metadata and dashboard json data, 
+  organized by directories'''
+  map = {}
+
+  for path in paths:
+    pathParts = path.split('/')
+
+    # filter out non-dashboards files
```
EvanSimpson, line 33: Can we also filter out anything beyond depth of 1 below `dashboards/` ?
For example, there is a `deprecated/` folder under MongoDB. It would be fine, and probably preferable to skip the validation for those files.

so
```python
if not 'dashboards' == pathParts[1] or len(pathParts) > 4:
  continue
```

scripts/dashboard/validate_dashboards_format.py
```diff
@@ -0,0 +1,153 @@
...
+  '''Checks if file is in json format and returns data in dictionary form'''
+  with open(path) as f:
+    try:
+      dashboards_dict = json.load(f)
+      dashboards_json = json.dumps(dashboards_dict)
+
+      return dashboards_dict
+
+    except JsonFormattingError:
+      print("{} content could not be loaded".format(path))
```
EvanSimpson, line 61: I think it would be appropriate to just have the error throw here.

EvanSimpson (changes requested):

scripts/dashboard/validate_dashboards_format.py
```diff
@@ -0,0 +1,153 @@
...
+
+class MetadataInJsonMisMatchError(Exception):
+  '''Raised when data in metadata file does not have matching data in a json file'''
+  pass
+
+class JsonInMetadataMisMatchError(Exception):
+  '''Raised when data in json file does not have matching data in a metadata file'''
+  pass
+
+def mapPaths(paths):
```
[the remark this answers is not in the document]
philshin-google: done.

scripts/dashboard/validate_dashboards_format.py
```diff
@@ -0,0 +1,153 @@
...
+  if not sample_dashboards:
+    raise MetadataFormattingError("sample_dashboards not defined in {}".format(path))
+  for sample_dashboard in sample_dashboards:
+    # Call function to check metadata file fields
+    check_metadata_entries(path, sample_dashboard)
+
+    dashboardId = sample_dashboard.get('id')
+    dashboardName = sample_dashboard.get('display_name')
+
+    pathMap[directory]['metadataIds'][dashboardId]=''
```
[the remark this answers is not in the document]
philshin-google: done.

scripts/dashboard/validate_dashboards_format.py
philshin-google: done.

scripts/dashboard/validate_dashboards_format.py
philshin-google: done in docstring for parent function.

scripts/dashboard/validate_dashboards_format.py
philshin-google: done. created refactored function for json validation check and raised the error directly based on return of that refactored check function.

scripts/dashboard/validate_dashboards_format.py
philshin-google: done.

scripts/dashboard/validate_dashboards_format.py
philshin-google: done.

scripts/dashboard/validate_dashboards_format.py
philshin-google: done.

scripts/dashboard/validate_dashboards_format.py
```diff
@@ -0,0 +1,168 @@
...
+      map[path_parts[2]]['json_files'].append(path_parts[3])
+
+  return map
+
+def get_sample_dashboards_json(path):
+  '''Checks if file is in json format and returns data in dictionary form'''
+  with open(path) as f:
+    dashboards_dict = json.load(f)
+
+    if not validate_json(dashboards_dict):
```
EvanSimpson, line 64: I think json.load(f) is what's going to fail and throw here. If .load() is successful, would .dumps() fail?

philshin-google: yeah I think you're right. Does that mean json.dumps was useless in the alerts checks I originally based this script off of? It seemed like the only reason they used json.dumps was for a check.

I can move the json.load method to the validation function.

philshin-google: done.

mr-arc (approved):

PR #342, opened 2022-11-01
Add single quota alert policies to google-quotas

rheewalt:
Screen Shot 2022-11-01 at 11 51 20 AM
Screen Shot 2022-11-01 at 11 39 18 AM
Screen Shot 2022-11-01 at 11 34 25 AM
Screen Shot 2022-11-01 at 9 07 19 AM

### DIFF
--- a/alerts/google-quotas/metadata.yaml
+++ b/alerts/google-quotas/metadata.yaml
@@ -4,4 +4,23 @@ alert_policy_templates:
   display_name: Quota Exceeded Alert Policy
   description: "Alert policy for exceeding quota limit"
   version: 1
-
+-
+  id: single-allocation-quota
+  display_name: High quota utilization
+  description: "Monitoring allocation quota utilization and will notify you if the utilization rises above alert trigger threshold."
+  version: 1
+-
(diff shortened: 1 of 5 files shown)


alerts/google-quotas/metadata.yaml
```diff
@@ -4,4 +4,23 @@ alert_policy_templates:
   display_name: Quota Exceeded Alert Policy
   description: "Alert policy for exceeding quota limit"
   version: 1
-
+-
+  id: single-allocation-quota
+  display_name: High quota utilization
+  description: "Monitoring quota utilization and will notify you if the utilization rises above alert trigger threshold."
```
cocosheng, line 10: the description for the newly added templates are all the same currently. would it be easier to understand if more details are included so users can differentiate?

cocosheng (changes requested):

alerts/google-quotas/metadata.yaml
rheewalt: Thanks.  Yes.  I've updated the description to match the templates they are describing.

PR #343, opened 2022-11-09
Add Prometheus GMP documentation

jsirianni:
This PR adds documentation configuration for the Redis GMP integration.

- Added redis documentation.yaml
- Added redis prometheus_metadata.yaml

Alerting has not been implemented yet. The alerting fields have been commented for now. Follow up PRs will include alert configuration and documentation.

### DIFF
--- a/dashboards/redis/documentation.yaml
+++ b/dashboards/redis/documentation.yaml
@@ -1,0 +1,62 @@
+exporter_type: sidecar
+app_name_short: Redis
+app_name: {{app_name_short}}
+app_site_name: Redis
+app_site_url: https://www.redis.com/
+exporter_name: Redis
+exporter_pkg_name: redis_exporter
+exporter_repo_url: https://github.com/oliver006/redis_exporter
+dashboard_available: true
+minimum_exporter_version: v1.43.1
+multiple_dashboards: false
+dashboard_display_name: {{app_name_short}} Prometheus Overview
+config_mods: |
+  apiVersion: apps/v1
+  kind: StatefulSet
+  metadata:
+    name: redis
+  spec:
+    serviceName: redis
+    selector:
+      matchLabels:
+  +     app.kubernetes.io/name: redis
+    template:
+      metadata:
+        labels:
+  +       app.kubernetes.io/name: redis
+      spec:
+        containers:
+        - name: redis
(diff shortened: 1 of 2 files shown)


dashboards/redis/prometheus_metadata.yaml
```diff
@@ -0,0 +1,32 @@
...
+        value_type: DOUBLE
+      - name: prometheus.googleapis.com/redis_memory_used_bytes/gauge
+        prometheus_name: redis_memory_used_bytes
+        kind: GAUGE
+        value_type: DOUBLE
+      - name: prometheus.googleapis.com/redis_net_output_bytes_total/counter
+        prometheus_name: redis_net_output_bytes_total
+        kind: CUMULATIVE
+        value_type: DOUBLE
+    install_documentation_url: https://cloud.google.com/stackdriver/docs/managed-prometheus/exporters
```
EvanSimpson, line 32:
```suggestion
    install_documentation_url: https://cloud.google.com/stackdriver/docs/managed-prometheus/exporters/redis
```

EvanSimpson (changes requested):

dashboards/redis/documentation.yaml
```diff
@@ -0,0 +1,62 @@
...
+  apiVersion: monitoring.googleapis.com/v1
+  kind: PodMonitoring
+  metadata:
+    name: redis
+    labels:
+      app.kubernetes.io/name: redis
+      app.kubernetes.io/part-of: google-cloud-managed-prometheus
+  spec:
+    endpoints:
+    - port: prometheus
```
yqlu, line 53: In our original iteration, we used a numbered port, but agree that going forward named ports are preferable when possible.

yqlu (approved):
Cool, since redis is one of the first wave (with nginx, istio, kube-state-metrics, node-exporter) that we already worked on, I could review the diff of this vs vs the sample one mary sent over. 

Some small comments but LGTM

dashboards/redis/documentation.yaml
jsirianni: Thanks for the review. On our end, we are trying to be as consistent as possible with things like this.

EvanSimpson (approved):

PR #344, opened 2022-11-09
Add related integrations to an alert template

turnerhayes:
Starting with one integration for now; will add others as needed

### DIFF
--- a/alerts/activemq/metadata.yaml
+++ b/alerts/activemq/metadata.yaml
@@ -5,12 +5,18 @@ alert_policy_templates:
   description: "When the broker's disk storage is near its limit(90%), then the producer
     flow control may cause producers to stop sending messages."
   version: 1
+  related_integrations:
+    - id: activemq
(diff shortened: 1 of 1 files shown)


EvanSimpson (approved):

alerts/activemq/metadata.yaml
```diff
@@ -5,16 +5,25 @@ alert_policy_templates:
   description: "When the broker's disk storage is near its limit(90%), then the producer
     flow control may cause producers to stop sending messages."
   version: 1
+  related_integrations:
```
cocosheng, line 8: @EvanSimpson would it be easier to move this field on top of alert_policy_templates? i assume it would apply to all templates.

EvanSimpson: We should do this on a per-template basis, to cover the possibility of multiple versions of a single integration requiring distinct alert templates.

PR #350, opened 2022-11-14
GMP Documentation: ActiveMQ

jsirianni:
This PR adds documentation configuration for the ActiveMQ GMP integration.

- Added documentation.yaml
- Added prometheus_metadata.yaml

### DIFF
--- a/integrations/activemq/documentation.yaml
+++ b/integrations/activemq/documentation.yaml
@@ -124,5 +124,4 @@ podmonitoring_config: |
 additional_prereq_info: |
   ActiveMQ Brokers can be configured for [JMX](https://activemq.apache.org/jmx){:class=external} 
   by setting the `ACTIVEMQ_JMX` and `ACTIVEMQ_OPTS` environment variables.
-additional_alert_info: You can adjust the alert thresholds to suit your application.
 sample_promql_query: up{job="activemq", cluster="{{cluster_name}}", namespace="{{namespace_name}}"}
--- a/integrations/activemq/documentation.yaml
+++ b/integrations/activemq/documentation.yaml
@@ -3,7 +3,7 @@ app_name_short: ActiveMQ
 app_name: {{app_name_short}}
 app_site_name: ActiveMQ
 app_site_url: https://activemq.apache.org/
-exporter_name: the ActiveMQ exporter
+exporter_name: the JMX exporter
 exporter_pkg_name: jmx-exporter
 exporter_repo_url: https://github.com/bitnami/containers/tree/main/bitnami/jmx-exporter#readme
 dashboard_available: true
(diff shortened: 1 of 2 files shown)


integrations/activemq/prometheus_metadata.yaml
```diff
@@ -0,0 +1,56 @@
+platforms:
+  - type: GKE
+    launch_stage: GA
+    detections:
+      - characteristic_metric:
+          metric_type: prometheus.googleapis.com/activemq_connections_total/gauge
+    exporter_metadata:
+      name: ActiveMQ Prometheus Exporter
```
varun-c, line 8: Based on the URL, "JMX Exporter" feels like a more appropriate name for this exporter, unless the doc url is for the wrong exporter. The bitnami exporter also does not seem to have a version 0.17.0. Did we mean to use this instead: https://github.com/prometheus/jmx_exporter?

varun-c (commented):
prometheus_metadata.yaml looks good to me, except for the one comment. Adding YQ to review documentation.yaml.

integrations/activemq/prometheus_metadata.yaml
jsirianni: I think "JMX Exporter" makes sense. A handful of integrations rely on it. I will update.

Bitnami packages `prometheus/jmx_exporter` into a container image, and it does have tag `0.17.0`. The Bitnami readme links to the Prometheus repo. We can switch to the Prometheus link if you think that makes more sense.

integrations/activemq/documentation.yaml
```diff
@@ -0,0 +1,127 @@
...
+  +     volumes:
+  +     - name: activemq-exporter
+  +       configMap:
+  +         name: activemq-exporter
+  +         items:
+  +         - key: config.yaml
+  +           path: config.yaml
+additional_install_info: |
+  These instructions assume you already have a working {{app_name_short}}
+  installation and want to modify it to include an exporter.
```
yqlu, line 106: Given that the template will already say "To install {{exporter_literal}} as a sidecar to {{app_name_short}}, modify your {{app_name_short}} configuration as shown in the following
example:"

Maybe this is redundant. 

Let's make the standard that we add a clarifying paragraph like this only when (as a bonus, as with the Redis case) followed immediately by additional detail for the other use case ("To set up {{app_name_short}} and {{exporter_name}} together from scratch, use this Helm chart...").

yqlu (approved):

integrations/activemq/prometheus_metadata.yaml
varun-c: We can keep the url as is if the bitnami jmx exporter is what we want users to be installing.

varun-c (approved):

yqlu (approved):

PR #366, opened 2022-11-21
GMP Documentation and alerts: Kafka 

jsirianni:
This PR adds documentation configuration for the Kafka GMP integration.

-  Added documentation.yaml

Documentation yaml includes alerts ported from https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/blob/master/alerts/kafka-gke/kafka-change-in-number-of-isrs.v1.json.

### DIFF
--- a/integrations/kafka/documentation.yaml
+++ b/integrations/kafka/documentation.yaml
@@ -1,4 +1,4 @@
-exporter_type: generic_standalone
+exporter_type: standalone
 app_name_short: Kafka
 app_name: {{app_name_short}}
 app_site_name: Kafka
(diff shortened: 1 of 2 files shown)


integrations/kafka/documentation.yaml
```diff
@@ -0,0 +1,96 @@
...
+    - port: prometheus
+      scheme: http
+      interval: 30s
+      path: /metrics
+    selector:
+      matchLabels:
+        app.kubernetes.io/name: kafka-exporter
+additional_install_info: |
+  The {{app_name_short}} exporter is configurable with [flags](https://github.com/danielqsj/kafka_exporter#flags){:class=external},
+  which can be set as container args. This example assumes Kafka is available via a ClusterIP service named `kafka`
```
mtabasko, line 57: would "as" work here? (We're not supposed to use Latin!)

integrations/kafka/documentation.yaml
```diff
@@ -0,0 +1,96 @@
+exporter_type: generic_standalone
```
mtabasko, line 1: What is "generic_standalone" supposed to convey? The only values we currently support are "standalone", "sidecar", "included", and "infrastructure"? Does this need to be handled specially, or was it intended to be "standalone"?

mtabasko (changes requested):

yqlu (approved):

integrations/kafka/documentation.yaml
jsirianni: Standalone is correct, our mistake.

mtabasko (approved):

EvanSimpson (approved):

PR #368, opened 2022-11-21
GMP Documentation: Consul

jsirianni:
This PR adds documentation configuration for the Consul GMP integration.

-  Added documentation.yaml

### DIFF
--- a/integrations/consul/documentation.yaml
+++ b/integrations/consul/documentation.yaml
@@ -1,0 +1,89 @@
+exporter_type: included
+app_name_short: Consul
+app_name: {{app_name_short}}
+app_site_name: Consul
+app_site_url: https://www.consul.io/
+exporter_name: the Consul exporter
+exporter_pkg_name: consul
+exporter_repo_url: https://developer.hashicorp.com/consul/docs/agent/config/config-files#telemetry-parameters
+dashboard_available: true
+minimum_exporter_version: "1.13.3"
+multiple_dashboards: false
+dashboard_display_name: {{app_name_short}} Prometheus Overview
+config_mods: |
+  apiVersion: v1
+  kind: ConfigMap
+  metadata:
+    name: consul-telemetry
+  data:
+  + metrics.hcl: |
+  +   telemetry {
+  +     prometheus_retention_time = "60s"
+  +     disable_hostname = true
+  +   }
+  ---
+  apiVersion: apps/v1
+  kind: StatefulSet
+  metadata:
+    name: consul
+  spec:
+    serviceName: consul
+    selector:
+      matchLabels:
+  +     app.kubernetes.io/name: consul
+    template:
+      metadata:
+        labels:
+  +       app.kubernetes.io/name: consul
+      spec:
+        containers:
+        - name: consul
+          image: consul:1.13.3
+          ports:
+          - containerPort: 8500
+            name: consul
+            protocol: TCP
+  +       volumeMounts:
+  +       - mountPath: /consul/config/metrics.hcl
+  +         subPath: metrics.hcl
+  +         name: consul-telemetry
+  +     volumes:
+  +     - name: consul-telemetry
+  +       configMap:
+  +         name: consul-telemetry
+  +         items:
+  +         - key: metrics.hcl
+  +           path: metrics.hcl
+podmonitoring_config: |
+  apiVersion: monitoring.googleapis.com/v1
+  kind: PodMonitoring
+  metadata:
+    name: consul
+    labels:
+      app.kubernetes.io/name: consul
+      app.kubernetes.io/part-of: google-cloud-managed-prometheus
+  spec:
+    endpoints:
+  + - port: 8500
+      scheme: http
+      interval: 30s
+  +   path: /v1/agent/metrics
+  +   params:
+  +     format: 
+  +       - prometheus
+    selector:
+      matchLabels:
+  +     app.kubernetes.io/name: consul
+additional_prereq_info: |
+  {{app_name_short}} exposes Prometheus-format metrics when the telemetry configuration options
+  `prometheus_retention_time` and `disable_hostname` are set. These options can be set by mounting
+  a configmap at `/consul/config/metrics.hcl`. Hashicorp recommends that a retention time twice 
+  that of the collection interval be used.
+
+   *  To determine if {{app_name_short}} is exposing metrics, run the following command:
+
+      <pre class="devsite-click-to-copy">
+      kubectl -n <namespace_name> port-forward <consul_pod_name> 8500
+      curl localhost:8500/v1/agent/metrics?format=prometheus
+      </pre>
+sample_promql_query: up{job="consul", cluster="{{cluster_name}}", namespace="{{namespace_name}}"}
--- a/integrations/consul/documentation.yaml
+++ b/integrations/consul/documentation.yaml
@@ -70,7 +70,7 @@ podmonitoring_config: |
       path: /v1/agent/metrics
       params:
         format: 
-          - prometheus
+        - prometheus
     selector:
       matchLabels:
         app.kubernetes.io/name: consul
--- a/integrations/consul/documentation.yaml
+++ b/integrations/consul/documentation.yaml
@@ -80,7 +80,7 @@ additional_prereq_info: |
   a configmap at `/consul/config/metrics.hcl`. Hashicorp recommends using a retention time twice that
   of the collection interval.
 additional_install_info: |
-  To determine if {{app_name_short}} is exposing metrics, run the following commands:
+  To determine if {{app_name_short}} is exposing metrics, run the following command:
 
   <pre class="devsite-click-to-copy">
   kubectl exec -n {{namespace_name}} {{pod_name}} -- curl -sS localhost:8500/v1/agent/metrics?format=prometheus
--- a/integrations/consul/documentation.yaml
+++ b/integrations/consul/documentation.yaml
@@ -80,10 +80,9 @@ additional_prereq_info: |
   a configmap at `/consul/config/metrics.hcl`. Hashicorp recommends using a retention time twice that
   of the collection interval.
 additional_install_info: |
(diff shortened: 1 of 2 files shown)


integrations/consul/documentation.yaml
```diff
@@ -0,0 +1,89 @@
...
+  +   params:
+  +     format: 
+  +       - prometheus
+    selector:
+      matchLabels:
+  +     app.kubernetes.io/name: consul
+additional_prereq_info: |
+  {{app_name_short}} exposes Prometheus-format metrics when the telemetry configuration options
+  `prometheus_retention_time` and `disable_hostname` are set. These options can be set by mounting
+  a configmap at `/consul/config/metrics.hcl`. Hashicorp recommends that a retention time twice
```
mtabasko, line 80: using

integrations/consul/documentation.yaml
```diff
@@ -0,0 +1,89 @@
...
+  +     format: 
+  +       - prometheus
+    selector:
+      matchLabels:
+  +     app.kubernetes.io/name: consul
+additional_prereq_info: |
+  {{app_name_short}} exposes Prometheus-format metrics when the telemetry configuration options
+  `prometheus_retention_time` and `disable_hostname` are set. These options can be set by mounting
+  a configmap at `/consul/config/metrics.hcl`. Hashicorp recommends that a retention time twice 
+  that of the collection interval be used.
```
mtabasko, line 81: if you say "Hashicorp recommends using... " above, you can delete "be used." here, so the sentence reads

Hashicorp recommends using a retention time twice that of the collection interval.

integrations/consul/documentation.yaml
```diff
@@ -0,0 +1,89 @@
...
+
+   *  To determine if {{app_name_short}} is exposing metrics, run the following command:
+
+      <pre class="devsite-click-to-copy">
+      kubectl -n <namespace_name> port-forward <consul_pod_name> 8500
+      curl localhost:8500/v1/agent/metrics?format=prometheus
+      </pre>
```
mtabasko, line 88: I'd just make this a paragraph. (One item isn't really a list.) Actually, I think I'd move this "To determine" bit to "additional_install_info" -- if I understand correctly, you won't see anything until you've made the configuration mods, so asking them to do this before they have made the config mods (as a "prereq") probably doesn't make sense. (But I may not be understanding this correctly!)

mtabasko (changes requested):

integrations/consul/documentation.yaml
```diff
@@ -0,0 +1,89 @@
...
+    selector:
+      matchLabels:
+  +     app.kubernetes.io/name: consul
+additional_prereq_info: |
+  {{app_name_short}} exposes Prometheus-format metrics when the telemetry configuration options
+  `prometheus_retention_time` and `disable_hostname` are set. These options can be set by mounting
+  a configmap at `/consul/config/metrics.hcl`. Hashicorp recommends that a retention time twice 
+  that of the collection interval be used.
+
+   *  To determine if {{app_name_short}} is exposing metrics, run the following command:
```
mtabasko, line 83: the following commands:  (? this looks like a "kubectl" command followed by a "curl" command, unless somehow the "curl" is part of "kubectl"??)

integrations/consul/documentation.yaml
```diff
@@ -0,0 +1,89 @@
...
+  + - port: 8500
+      scheme: http
+      interval: 30s
+  +   path: /v1/agent/metrics
+  +   params:
+  +     format: 
+  +       - prometheus
+    selector:
+      matchLabels:
+  +     app.kubernetes.io/name: consul
```
yqlu, line 76: We've been using "+"s to indicate the delta to an existing config to add telemetry and monitoring, so this makes sense under config_mod, but we shouldn't need to have "+"s in podmonitoring_config since we don't expect anyone to have an existing podmonitoring config they are modifying.

integrations/consul/documentation.yaml
jsirianni: Hi @yqlu, I cannot access that link unfortunately.

I think I will take the second option. I try to avoid assuming curl will be available in the container image, but in consul's case it is available. 

Will the `{{ }}` syntax be expanded when the page is rendered?
```
kubectl exec -n {{namespace_name}} {{pod_name}}
```

integrations/consul/documentation.yaml
```diff
@@ -0,0 +1,88 @@
...
+    selector:
+      matchLabels:
+        app.kubernetes.io/name: consul
+additional_prereq_info: |
+  {{app_name_short}} exposes Prometheus-format metrics when the telemetry configuration options
+  `prometheus_retention_time` and `disable_hostname` are set. These options can be set by mounting
+  a configmap at `/consul/config/metrics.hcl`. Hashicorp recommends using a retention time twice that
+  of the collection interval.
+additional_install_info: |
+  To determine if {{app_name_short}} is exposing metrics, run the following commands:
```
mtabasko, line 83: Looks like you've made this one command, so "the following command" didn't need to be changed! (Sorry!!)

jsirianni: Oof, my bad. I went down the list of feedback and treated them independently :)

mtabasko (approved):
Thanks! Sorry for the nit-pickery!

integrations/consul/documentation.yaml
yqlu: Ah my bad, just fixed the link in my earlier comment. Yes, there is a macro that expands {{ }} to the editable variables

The code snippet in [the istio example](https://cloud.devsite.corp.google.com/stackdriver/docs/managed-prometheus/exporters/istio) looks like 
```
<pre class="devsite-click-to-copy">
kubectl exec {{pod_name}} -n {{namespace_name}} -c istio-proxy -- curl -sS 'localhost:15090/stats/prometheus'
</pre>
```
in the underlying markdown

yqlu (approved):

integrations/consul/documentation.yaml
```diff
@@ -0,0 +1,88 @@
...
+      params:
+        format: 
+          - prometheus
+    selector:
+      matchLabels:
+        app.kubernetes.io/name: consul
+additional_prereq_info: |
+  {{app_name_short}} exposes Prometheus-format metrics when the telemetry configuration options
+  `prometheus_retention_time` and `disable_hostname` are set. These options can be set by mounting
+  a configmap at `/consul/config/metrics.hcl`. Hashicorp recommends using a retention time twice that
```
mtabasko, line 80: HashiCorp, not Hashicorp

integrations/consul/documentation.yaml
```diff
@@ -0,0 +1,88 @@
+exporter_type: included
+app_name_short: Consul
+app_name: {{app_name_short}}
```
mtabasko, line 3: app_name: HashiCorp {{app_name_short}}

mtabasko:
Could I make two tiny last-minute requests, since this hasn't gone it yet?? This is entirely my fault for not catching this, but 
I completely neglected to check whether "app_name" should be more specific than "app_name_short" (also missed this on ActiveMQ, CouchDB, and Zookeeper, all of which should be Apache {{app_name_short}}. :( Deepest apologies for my loserdom....

Line 3: app_name: HashiCorp {{app_name_short}}
on Line 80: Hashicorp -> HashiCorp

integrations/consul/documentation.yaml
jsirianni: > 

No problem, your feedback is appreciated. I will make a quick PR for the others.

mtabasko (approved):
Thank you, Joe!

integrations/consul/prometheus_metadata.yaml
```diff
@@ -0,0 +1,56 @@
...
+        value_type: DOUBLE
+      - name: prometheus.googleapis.com/consul_raft_state_candidate/counter
+        prometheus_name: consul_raft_state_candidate
+        kind: CUMULATIVE
+        value_type: DOUBLE
+      - name: prometheus.googleapis.com/consul_raft_state_leader/counter
+        prometheus_name: consul_raft_state_leader
+        kind: CUMULATIVE
+        value_type: DOUBLE
+    install_documentation_url: https://cloud.google.com/stackdriver/docs/managed-prometheus/exporters
```
EvanSimpson, line 56:
```suggestion
    install_documentation_url: https://cloud.google.com/stackdriver/docs/managed-prometheus/exporters/consul
```

EvanSimpson (changes requested):

integrations/consul/documentation.yaml
```diff
@@ -0,0 +1,88 @@
...
+      app.kubernetes.io/part-of: google-cloud-managed-prometheus
+  spec:
+    endpoints:
+    - port: 8500
+      scheme: http
+      interval: 30s
+      path: /v1/agent/metrics
+      params:
+        format: 
+          - prometheus
```
mtabasko, line 73: Our YAML linter complained that this line, -prometheus, was indented 2 spaces too many.

mtabasko (changes requested):
Our linter complained about line 73 of "documentation.yaml" -- indented 2 spaces to many.

EvanSimpson (approved):

integrations/consul/documentation.yaml
```diff
@@ -0,0 +1,88 @@
...
+  apiVersion: monitoring.googleapis.com/v1
+  kind: PodMonitoring
+  metadata:
+    name: consul
+    labels:
+      app.kubernetes.io/name: consul
+      app.kubernetes.io/part-of: google-cloud-managed-prometheus
+  spec:
+    endpoints:
+    - port: 8500
```
yqlu, line 67: Just double-checking, can this be a named port?

mtabasko (approved):

integrations/consul/documentation.yaml
jsirianni: We did not use Helm or an Operator to deploy consul, I think it is okay to assume 8500 as this is the default port for Consul.

PR #371, opened 2022-11-29
GMP Documentation and alert: Elasticsearch

jsirianni:
This PR adds documentation configuration for the Elasticsearch GMP integration.

-  Added documentation.yaml

The documentation yaml includes an alert ported from https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/tree/master/alerts/elasticsearch-gke

### DIFF
--- a/integrations/elasticsearch/documentation.yaml
+++ b/integrations/elasticsearch/documentation.yaml
@@ -51,11 +51,11 @@ podmonitoring_config: |
     selector:
       matchLabels:
         app.kubernetes.io/name: elasticsearch-exporter
-additional_prereq_info: |
+additional_install_info: |
   The {{app_name_short}} exporter is configurable with [flags](https://github.com/prometheus-community/elasticsearch_exporter#configuration){:class=external},
   which can be set as container args. You can adjust your exporter's configuration to suit your {{app_name_short}} configuration.
 
-  Update the `--es.uri` flag to match your envirionments username, password, and clusterIP service name.
+  Update the `--es.uri` flag to match your envirionment's username, password, and clusterIP service name.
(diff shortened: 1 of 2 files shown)


integrations/elasticsearch/prometheus_metadata.yaml
```diff
@@ -0,0 +1,69 @@
+platforms:
+  - type: GKE
+    launch_stage: GA
+    exporter_metadata:
```
EvanSimpson, line 4: Can you add a characteristic metric detection as well?

EvanSimpson (changes requested):

yqlu (approved):
👍  in terms of validating the config on a cluster myself

Tagging in @mtabasko to review the docs format. This is probably significant in the sense that it is our first "standalone" type exporter doc, so we want to ensure the template works.

integrations/elasticsearch/documentation.yaml
```diff
@@ -0,0 +1,107 @@
...
+  spec:
+    endpoints:
+    - port: prometheus
+      scheme: http
+      interval: 30s
+      path: /metrics
+    selector:
+      matchLabels:
+        app.kubernetes.io/name: elasticsearch-exporter
+additional_prereq_info: |
```
mtabasko, line 54: This seems like "additional_install_info" (it makes sense after you've seen the install config).

mtabasko:
Other than the one small change, from "additional_prereq_info" to "additional_install_info", I don't see any problems. (We had actually done a version of Elasticsearch as a "standalone" exporter when building the template originally -- the biggest diff between "standalone" and "sidecar" in terms of the template is whether we use "config_mods" or "standalone_config" for the config info, and that all looks good here to me! Thanks.

mtabasko (approved):
docs look good. Thanks again!

EvanSimpson (approved):

PR #374, opened 2022-11-30
GMP Documentation and alert: Ingress NGINX

jsirianni:
This PR adds documentation configuration for the Ingress NGINX GMP integration.

-  Added documentation.yaml

The documentation yaml includes an alert ported from https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/tree/master/alerts/nginx-ingress-gke

### DIFF
--- a/integrations/ingress-nginx/documentation.yaml
+++ b/integrations/ingress-nginx/documentation.yaml
@@ -11,12 +11,11 @@ additional_prereq_info: |
   have to install it separately. You can run the following checks to verify
   that {{app_name_short}} is exposing Prometheus-format metrics.
 
-   *  To determine if {{exporter_name}} is exposing metrics, run the following command:
+  To determine if {{exporter_name}} is exposing metrics, run the following command:
 
-      <pre class="devsite-click-to-copy">
-      kubectl -n ingress-nginx port-forward deploy/ingress-nginx-controller 10254
-      curl localhost:10254/metrics
-      </pre>
+  <pre class="devsite-click-to-copy">
+  kubectl exec -n {{ namespace }} deploy/ingress-nginx-controller -- curl http://localhost:10254/metrics
+  </pre>
 dashboard_available: true
 multiple_dashboards: false
 dashboard_display_name: {{app_name_short}} Prometheus Overview
@@ -30,13 +29,13 @@ podmonitoring_config: |
       app.kubernetes.io/part-of: google-cloud-managed-prometheus
   spec:
     endpoints:
-  + - port: 10254
+    - port: 10254
       scheme: http
       interval: 30s
       path: /metrics
     selector:
       matchLabels:
-  +     app.kubernetes.io/name: ingress-nginx
+        app.kubernetes.io/name: ingress-nginx
 additional_podmonitoring_info: |
   Ensure that the port and matchLabels match that of the {{app_name_short}} pods you wish to monitor. By default,
   {{app_name_short}} exposes metrics on port 10254 and has the label `app.kubernetes.io/name: ingress-nginx`.
--- a/integrations/ingress-nginx/documentation.yaml
+++ b/integrations/ingress-nginx/documentation.yaml
@@ -1,0 +1,91 @@
+exporter_type: included
+app_name_short: Ingress NGINX
+app_name: {{app_name_short}}
+app_site_name: Ingress NGINX
+app_site_url: https://kubernetes.github.io/ingress-nginx
+exporter_name: the Ingress NGINX exporter
+exporter_pkg_name: ingress-nginx
+exporter_repo_url: https://kubernetes.github.io/ingress-nginx/user-guide/monitoring/
+additional_prereq_info: |
+  {{app_name_short}} exposes Prometheus-format metrics automatically; you do not
+  have to install it separately. You can run the following checks to verify
+  that {{app_name_short}} is exposing Prometheus-format metrics.
+
+   *  To determine if {{exporter_name}} is exposing metrics, run the following command:
+
+      <pre class="devsite-click-to-copy">
+      kubectl -n ingress-nginx port-forward deploy/ingress-nginx-controller 10254
+      curl localhost:10254/metrics
+      </pre>
+dashboard_available: true
+multiple_dashboards: false
+dashboard_display_name: {{app_name_short}} Prometheus Overview
+podmonitoring_config: |
+  apiVersion: monitoring.googleapis.com/v1
+  kind: PodMonitoring
+  metadata:
+    name: ingress-nginx
+    labels:
+      app.kubernetes.io/name: ingress-nginx
+      app.kubernetes.io/part-of: google-cloud-managed-prometheus
+  spec:
+    endpoints:
+  + - port: 10254
+      scheme: http
+      interval: 30s
+      path: /metrics
+    selector:
+      matchLabels:
+  +     app.kubernetes.io/name: ingress-nginx
+additional_podmonitoring_info: |
+  Ensure that the port and matchLabels match that of the {{app_name_short}} pods you wish to monitor. By default,
+  {{app_name_short}} exposes metrics on port 10254 and has the label `app.kubernetes.io/name: ingress-nginx`.
+sample_promql_query: up{job="ingress-nginx", cluster="{{cluster_name}}", namespace="{{namespace_name}}"}
+alerts_config: |
+  apiVersion: monitoring.googleapis.com/v1
+  kind: Rules
+  metadata:
+    name: nginx-ingress-rules
+    labels:
+      app.kubernetes.io/component: rules
+      app.kubernetes.io/name: nginx-ingress-rules
+      app.kubernetes.io/part-of: google-cloud-managed-prometheus
+  spec:
+    groups:
+    - name: nginx-ingress
+      interval: 30s
+      rules:
+      - alert: NGINXIngressDroppedConnections
+        annotations:
+          description: |-
+            NGINX Ingress dropped connections
+              VALUE = {{ $value }}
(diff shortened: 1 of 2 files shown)


integrations/ingress-nginx/documentation.yaml
```diff
@@ -0,0 +1,90 @@
...
+    endpoints:
+    - port: 10254
+      scheme: http
+      interval: 30s
+      path: /metrics
+    selector:
+      matchLabels:
+        app.kubernetes.io/name: ingress-nginx
+additional_podmonitoring_info: |
+  Ensure that the port and matchLabels match that of the {{app_name_short}} pods you wish to monitor. By default,
```
mtabasko, line 40: the values of the `port` and `matchLabels` fields match those  ... you want to monitor.

(We try to avoid "wishing" in docs. If you could make things work by just wishing, life would be soooo much easier!)

integrations/ingress-nginx/documentation.yaml
```diff
@@ -0,0 +1,90 @@
...
+  {{app_name_short}} exposes metrics on port 10254 and has the label `app.kubernetes.io/name: ingress-nginx`.
```
mtabasko, line 41: uses?  (it look me a while to figure out that this was referring back to the "values you want to match" -- I was reading it as "app_name_short (a) exposes metrics on port 10254 and (b) has the label .." but it's the config that has the label, not app_name_short.)

integrations/ingress-nginx/documentation.yaml
```diff
@@ -0,0 +1,90 @@
...
+app_site_url: https://kubernetes.github.io/ingress-nginx
+exporter_name: the Ingress NGINX exporter
+exporter_pkg_name: ingress-nginx
+exporter_repo_url: https://kubernetes.github.io/ingress-nginx/user-guide/monitoring/
+additional_prereq_info: |
+  {{app_name_short}} exposes Prometheus-format metrics automatically; you do not
+  have to install it separately. You can run the following checks to verify
+  that {{app_name_short}} is exposing Prometheus-format metrics.
+
+  To determine if {{exporter_name}} is exposing metrics, run the following command:
```
mtabasko, line 14: You could make this (lines 11-14) shorter by just saying:

...separately. To verify that {{app_name_short}} is exposing metrics, run the following command:

integrations/ingress-nginx/documentation.yaml
```diff
@@ -0,0 +1,90 @@
+exporter_type: included
+app_name_short: Ingress NGINX
+app_name: {{app_name_short}}
```
mtabasko, line 3: Should this maybe be "{app_name_short}} Controller" ?

The first line on the webpage we're linking to says "This is the documentation for the Ingress NGINX Controller."

integrations/ingress-nginx/documentation.yaml
```diff
@@ -0,0 +1,90 @@
...
+app_site_name: Ingress NGINX
```
mtabasko, line 4: It looks like the title of that web page is "NGINX Ingress Controller"

(would be nice if they were consistent in their names! I've seen "Ingress NGINX Controller", "NGINX Ingress Controller" and "Ingress-NGINX controller"!)

integrations/ingress-nginx/documentation.yaml
```diff
@@ -0,0 +1,90 @@
...
+app_site_url: https://kubernetes.github.io/ingress-nginx
+exporter_name: the Ingress NGINX exporter
```
mtabasko, line 6: should this maybe be "controller"? That seems to be how it's referred to.

mtabasko: if so, it might make the doc flow more naturally for people familiar with this if some of the occurrences of {{app_name_short}} in prose were replaced with {{exporter_name}}.

mtabasko (changes requested):

integrations/ingress-nginx/documentation.yaml
jsirianni: It is confusing. NGINX Ingress Controller is not the same as Ingress NGINX.  Ingress NGINX is the official ingress solution for Kubernetes, but other controllers exist such as haproxy and nginx ingress (from the nginx company, now F5). The core software is the same, but the featureset and configuration can differ. F5 offers paid support and features for NGINX Ingress (I believe), while Ingress NGINX is purely an open source solution provided by the Kubernetes project.

- Ingress NGINX: https://kubernetes.github.io/ingress-nginx
- NGINX Ingress: https://docs.nginx.com/nginx-ingress-controller/

This sheet has a nice breakdown, and turns out there are several other nginx controllers available from various vendors. https://docs.google.com/spreadsheets/d/191WWNpjJ2za6-nbG4ZoUMXMpUK8KlCIosvQB0f-oq3k/edit#gid=907731238

integrations/ingress-nginx/documentation.yaml
jsirianni: I might not be following 100%. When deploying ingress nginx using their official documentation, a deployment with the label `app.kubernetes.io/name: ingress-nginx` will be created. This does happen to match the label on the PodMonitoring configuration.

integrations/ingress-nginx/documentation.yaml
mtabasko: God, that is confusing. This value was just supposed to capture the title of the web page we're linking to (so they know they are in the right place). I hope that someone looking at this already knows what this is, so maybe it's only confusing to me. Would it help to make this "Kubernetes Ingress NGINX"? (I guess I don't understand why this page, a Kubernetes page, is titled NGINX Ingress Controller -- maybe that's the problem?) But I will defer to your judgement on this, because I'm hopelessly confused! Thanks for the background.

mtabasko: Also, does this mean we shouldn't change "app_name" (line 3) or "exporter_name" (line 6) to append "Controller"?? (That suggestion was inspired by the fact that the page at this link kept referring to "controller". But if that's not what we have here? (Again, if this was a stupid suggestion, please revert it -- as I said, I'm hopelessly confused about this at this point.)

mtabasko (changes requested):
Not so much "suggesting" changes as saying if my changes were stupid, please revert! :)

integrations/ingress-nginx/prometheus_metadata.yaml
```diff
@@ -0,0 +1,40 @@
...
+        value_type: DISTRIBUTION
+      - name: prometheus.googleapis.com/nginx_ingress_controller_nginx_process_resident_memory_bytes/gauge
+        prometheus_name: nginx_ingress_controller_nginx_process_resident_memory_bytes
+        kind: GAUGE
+        value_type: DOUBLE
+      - name: prometheus.googleapis.com/nginx_ingress_controller_nginx_process_cpu_seconds_total/counter
+        prometheus_name: nginx_ingress_controller_nginx_process_cpu_seconds_total
+        kind: CUMULATIVE
+        value_type: DOUBLE
+    install_documentation_url: https://cloud.google.com/stackdriver/docs/managed-prometheus/exporters/nginx-ingress
```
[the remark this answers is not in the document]
mtabasko: Roger that!

integrations/ingress-nginx/documentation.yaml
jsirianni: I think keeping "controller" makes sense. Yes, the ingress world is very confusing.

integrations/ingress-nginx/documentation.yaml
```diff
@@ -0,0 +1,88 @@
...
+app_name_short: Ingress NGINX
+app_name: {{app_name_short}} Controller
+app_site_name: Ingress NGINX
+app_site_url: https://kubernetes.github.io/ingress-nginx
+exporter_name: the Ingress NGINX Controller
+exporter_pkg_name: ingress-nginx
+exporter_repo_url: https://kubernetes.github.io/ingress-nginx/user-guide/monitoring/
+additional_prereq_info: |
+  {{app_name_short}} exposes Prometheus-format metrics automatically; you do not
+  have to install it separately. To verify that {{exporter_name}} is exposing
```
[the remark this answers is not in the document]
jsirianni: In my experience, we do not need to make changes to ingress nginx. The port 10254 is listening by default, and exposes metrics. This same port is used for healthchecks.

It may or may not be defined explicitly as a `containerPort`, but that does not prevent the port from being accessed by the prometheus scraper.

integrations/ingress-nginx/documentation.yaml
```diff
@@ -0,0 +1,88 @@
...
+  apiVersion: monitoring.googleapis.com/v1
+  kind: PodMonitoring
+  metadata:
+    name: ingress-nginx
+    labels:
+      app.kubernetes.io/name: ingress-nginx
+      app.kubernetes.io/part-of: google-cloud-managed-prometheus
+  spec:
+    endpoints:
+    - port: 10254
```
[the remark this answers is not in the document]
jsirianni: By default, the port is not named but I think we can assume that the port will be 10254, as this is the port used for healthchecks.

mtabasko (commented):
Current doc.yaml file looks OK to me, pending answers to technical questions about which I have nothing useful to say. One question about the "doc_url" in the prom_metadata.yaml file.

integrations/ingress-nginx/prometheus_metadata.yaml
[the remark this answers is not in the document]
mtabasko: It does seem like "ingress-nginx" is more consistent with the values in the "documentation.yaml" file, though.

mtabasko (approved):

mtabasko (approved):
Thanks again!

PR #384, opened 2022-12-02
GMP Dashboard filtering: Ingress NGINX

jsirianni:
Added `Cluster`, `Location`, and `Namespace` filtering capabilities.

### DIFF
--- a/dashboards/nginx-ingress/nginx-ingress-prometheus.json
+++ b/dashboards/nginx-ingress/nginx-ingress-prometheus.json
@@ -62,7 +62,7 @@
               "sparkChartType": "SPARK_LINE"
             },
             "timeSeriesQuery": {
-              "prometheusQuery": "sum(nginx_ingress_controller_success{controller_pod=~\".*\",controller_class=~\".*\",controller_namespace=~\".*\",${Cluster},${Location},${Namespace}})"
+              "prometheusQuery": "sum(nginx_ingress_controller_success{${Cluster},${Location},${Namespace}})"
             }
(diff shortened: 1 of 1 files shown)


dashboards/nginx-ingress/nginx-ingress-prometheus.json
```diff
@@ -44,7 +62,7 @@
               "sparkChartType": "SPARK_LINE"
             },
             "timeSeriesQuery": {
-              "prometheusQuery": "sum(nginx_ingress_controller_success{controller_pod=~\".*\",controller_class=~\".*\",controller_namespace=~\".*\"})"
+              "prometheusQuery": "sum(nginx_ingress_controller_success{controller_pod=~\".*\",controller_class=~\".*\",controller_namespace=~\".*\",${Cluster},${Location},${Namespace}})"
```
yqlu, line 65: These wildcard variables `controller_pod=~\".*\",controller_class=~\".*\",controller_namespace=~\".*\"` look vestigial -- should we clean them up? (Also applies to a few other queries in this dashboard)

Unless for ingress nginx specifically, it makes sense to add template variables to filter by these additional labels in addition to the standard cluster / location / namespace?

jsirianni: I agree.

And no, these values are not particularly important for filtering. I can see filtering on pod name being useful sometimes, primarily for statefulsets where the pod name is always the same.

yqlu (approved):

PR #388, opened 2022-12-02
Add GMP Rabbitmq doc

jsirianni:
This PR adds documentation configuration for the Rabbitmq GMP integration.

-  Added documentation.yaml

The documentation yaml includes an alert ported from https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/tree/master/alerts/rabbitmq-gke

### DIFF
--- a/integrations/rabbitmq/documentation.yaml
+++ b/integrations/rabbitmq/documentation.yaml
@@ -8,15 +8,17 @@ exporter_pkg_name: rabbitmq
 exporter_repo_url: https://rabbitmq.com/prometheus.html#overview-prometheus
 additional_prereq_info: |
   {{app_name_short}} exposes Prometheus-format metrics automatically; you do not
-  have to install it separately. You can run the following checks to verify
-  that {{app_name_short}} is exposing Prometheus-format metrics.
+  have to install it separately. To verify that {{exporter_name}} is emitting metrics on the expected
+  endpoints, set up port-forwarding with the following command:
 
-   *  To determine if {{exporter_name}} is exposing metrics, run the following command:
+  To determine if {{exporter_name}} is exposing metrics, run the following command:
 
-      <pre class="devsite-click-to-copy">
-      kubectl -n <namespace_name> port-forward <rabbitmq_pod_name> 15692
-      curl localhost:15692/metrics
-      </pre>
+  <pre class="devsite-click-to-copy">
+  kubectl -n {{namespace_name}} port-forward {{pod_name}} 15692
+  </pre>
+
+  Access the endpoint `localhost:15692/metrics` by using the browser or curl in another terminal session
+  to verify that the metrics are being exposed by the exporter for scraping.
 dashboard_available: true
 multiple_dashboards: false
 dashboard_display_name: {{app_name_short}} Prometheus Overview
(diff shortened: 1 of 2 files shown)


integrations/rabbitmq/prometheus_metadata.yaml
```diff
@@ -0,0 +1,128 @@
...
+  - type: GKE
+    detections:
+      - characteristic_metric:
+          metric_type: prometheus.googleapis.com/rabbitmq_connections/gauge
+    launch_stage: HIDDEN
+    exporter_metadata:
+      name: RabbitMQ Prometheus Exporter
+      doc_url: https://rabbitmq.com/prometheus.html#overview-prometheus
+      minimum_supported_version: v3.8.0
+    default_metrics:
```
varun-c, line 11: I noticed that the dashboard that uses this integration uses a rabbitmq_identity_info query. Is that something we should include in this list?

varun-c (commented):
The prometheus_metadata.yaml looks good. Will defer to YQ for the documentation.yaml file.

integrations/rabbitmq/documentation.yaml
```diff
@@ -0,0 +1,94 @@
...
+      scheme: http
+      interval: 30s
+      path: /metrics
+    selector:
+      matchLabels:
+        app.kubernetes.io/component: rabbitmq
+additional_podmonitoring_info: |
+  Ensure that the values of the port and matchLabels match that of the {{app_name_short}} pods you want to monitor.
+  Rabbitmq deployed by the [operator](https://www.rabbitmq.com/kubernetes/operator/operator-overview.html){:class=external}
+  will contain the label `app.kubernetes.io/component: rabbitmq` and `prometheus` port.
```
yqlu, line 45: This is great -- I tested these docs by installing RabbitMQ a different way (via [GCP click-to-deploy](https://github.com/GoogleCloudPlatform/click-to-deploy/tree/master/k8s/rabbitmq)) and it had [a different label](https://github.com/GoogleCloudPlatform/click-to-deploy/blob/master/k8s/rabbitmq/chart/rabbitmq/templates/statefulset.yaml#L7), but this was good enough to guide me to adapt the config accordingly.

integrations/rabbitmq/documentation.yaml
```diff
@@ -0,0 +1,95 @@
...
+    endpoints:
+    - port: prometheus
+      scheme: http
+      interval: 30s
+      path: /metrics
+    selector:
+      matchLabels:
+        app.kubernetes.io/component: rabbitmq
+additional_podmonitoring_info: |
+  Ensure that the values of the port and matchLabels match that of the {{app_name_short}} pods you want to monitor.
```
mtabasko, line 44: \`port\` and \`matchLabels\` fields match those of

(literal "port" and "matchLabels", add "fields", and there's more than one, so "match those", not "match that".

integrations/rabbitmq/documentation.yaml
```diff
@@ -0,0 +1,95 @@
...
+app_name: {{app_name_short}}
+app_site_name: RabbitMQ
+app_site_url: https://rabbitmq.com/
+exporter_name: the RabbitMQ exporter
+exporter_pkg_name: rabbitmq
+exporter_repo_url: https://rabbitmq.com/prometheus.html#overview-prometheus
+minimum_exporter_version: v3.8.0
+additional_prereq_info: |
+  {{app_name_short}} exposes Prometheus-format metrics automatically; you do not
+  have to install it separately. To verify that {{exporter_name}} is emitting metrics on the expected
```
mtabasko, line 12: This is a little misleading: To verify that ...., set up port-forwarding with the following command". Setting up port forwarding doesn't verify that the metrics are emitted -- it's part of it, but not the whole thing.

The line  "To determine if ..., run the following command" is similarly confusing -- that just sets up port forwarding.

What about something like this:

To verify that {{exporter_name}} is emitting metrics on the expected endpoints, do the following:

0. Set up port forwarding by using the following command: (lines 17-19)
0. Access the endpoint \`localhost:15692/metrics\` by using the browser or the \`curl\` utility in another terminal session.

integrations/rabbitmq/documentation.yaml
```diff
@@ -0,0 +1,95 @@
...
+  endpoints, set up port-forwarding with the following command:
+
+  To determine if {{exporter_name}} is exposing metrics, run the following command:
+
+  <pre class="devsite-click-to-copy">
+  kubectl -n {{namespace_name}} port-forward {{pod_name}} 15692
+  </pre>
+
+  Access the endpoint `localhost:15692/metrics` by using the browser or curl in another terminal session
```
mtabasko, line 21: the \`curl\` utility

integrations/rabbitmq/documentation.yaml
```diff
@@ -0,0 +1,95 @@
...
+  to verify that the metrics are being exposed by the exporter for scraping.
```
mtabasko, line 22: This is a bit redundant with lines 11-12. I think I'd just end this wtih "terminal session."

mtabasko (changes requested):

yqlu (approved):

mtabasko (approved):
Looks good to me. Thanks!!

PR #389, opened 2022-12-02
GMP NGINX doc and alerts configuration

jsirianni:
This doc and alerts is derived from the official doc [here](https://cloud.google.com/stackdriver/docs/managed-prometheus/exporters/nginx).

We were instructed to use launch stage "HIDDEN", however, I have left it on GA because the doc is already public. Let me know if this should change.

### DIFF
--- a/integrations/nginx/prometheus_metadata.yaml
+++ b/integrations/nginx/prometheus_metadata.yaml
@@ -45,4 +45,4 @@ platforms:
         prometheus_name: nginxexporter_build_info
         kind: GAUGE
         value_type: DOUBLE
-    install_documentation_url: https://cloud.google.com/stackdriver/docs/managed-prometheus/exporters/mysql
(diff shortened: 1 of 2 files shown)


integrations/nginx/prometheus_metadata.yaml
```diff
@@ -0,0 +1,48 @@
...
+        value_type: DOUBLE
+      - name: prometheus.googleapis.com/nginx_up/gauge
+        prometheus_name: nginx_up
+        kind: GAUGE
+        value_type: DOUBLE
+      - name: prometheus.googleapis.com/nginxexporter_build_info/gauge
+        prometheus_name: nginxexporter_build_info
+        kind: GAUGE
+        value_type: DOUBLE
+    install_documentation_url: https://cloud.google.com/stackdriver/docs/managed-prometheus/exporters/mysql
```
yqlu, line 48: this should be nginx

EvanSimpson (approved):

PR #390, opened 2022-12-02
GMP Istio doc and alerts configuration

jsirianni:
This doc and alerts is derived from the official doc [here](https://cloud.google.com/stackdriver/docs/managed-prometheus/exporters/istio).

We were instructed to use launch stage "HIDDEN", however, I have left it on GA because the doc is already public. Let me know if this should change.

### DIFF
--- a/integrations/istio/prometheus_metadata.yaml
+++ b/integrations/istio/prometheus_metadata.yaml
@@ -1,0 +1,63 @@
+platforms:
+  - type: GKE
+    detections:
+      - characteristic_metric:
+          metric_type: prometheus.googleapis.com/envoy_server_live/gauge
+    launch_stage: GA
+    exporter_metadata:
+      name: Istio Prometheus Exporter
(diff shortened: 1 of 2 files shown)


integrations/istio/prometheus_metadata.yaml
```diff
@@ -0,0 +1,63 @@
...
+        value_type: DOUBLE
+      - name: prometheus.googleapis.com/envoy_cluster_manager_cds_update_success/counter
+        prometheus_name: envoy_cluster_manager_cds_update_success
+        kind: CUMULATIVE
+        value_type: DOUBLE
+      - name: prometheus.googleapis.com/envoy_cluster_manager_cds_update_failure/counter
+        prometheus_name: envoy_cluster_manager_cds_update_failure
+        kind: CUMULATIVE
+        value_type: DOUBLE
+    install_documentation_url: https://cloud.google.com/stackdriver/docs/managed-prometheus/exporters/mysql
```
yqlu, line 63: This should be istio

PR #423, opened 2022-12-19
GMP Documentation, Alerts, and Filters: Velero

algchoo:
This PR includes:
* documentation
* alert definitions
* dashboard filters

### DIFF
--- a/integrations/velero/documentation.yaml
+++ b/integrations/velero/documentation.yaml
@@ -3,20 +3,23 @@ app_name_short: Velero
 app_name: {{app_name_short}}
 app_site_name: Velero
 app_site_url: https://velero.io/
-exporter_name: the elero exporter
+exporter_name: the Velero Prometheus exporter
 exporter_pkg_name: velero
 exporter_repo_url: https://velero.io/docs/
+minimum_exporter_version: v1.3.0
 additional_prereq_info: |
   {{app_name_short}} exposes Prometheus-format metrics automatically; you do not
-  have to install it separately. You can run the following checks to verify
-  that {{app_name_short}} is exposing Prometheus-format metrics.
+  have to install it separately. To verify that {{exporter_name}} is emitting
+  metrics on the expected endpoints, do the following:
 
-   *  To determine if {{app_name_short}} is exposing metrics, run the following command:
+  Set up port forwarding by using the following command:
 
-      <pre class="devsite-click-to-copy">
-      kubectl -n velero port-forward deploy/velero 8085
-      curl localhost:8085/metrics
-      </pre>
+  <pre class="devsite-click-to-copy">
+  kubectl -n velero port-forward deploy/velero 8085
+  </pre>
+
+  Access the endpoint `localhost:8085/metrics` by using the browser
+  or the `curl` utility in another terminal session.
 dashboard_available: true
 multiple_dashboards: false
 dashboard_display_name: {{app_name_short}} Prometheus Overview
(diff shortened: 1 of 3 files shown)


integrations/velero/documentation.yaml
```diff
@@ -0,0 +1,91 @@
...
+  {{app_name_short}} exposes Prometheus-format metrics automatically; you do not
+  have to install it separately. You can run the following checks to verify
+  that {{app_name_short}} is exposing Prometheus-format metrics.
+
+   *  To determine if {{app_name_short}} is exposing metrics, run the following command:
+
+      <pre class="devsite-click-to-copy">
+      kubectl -n velero port-forward deploy/velero 8085
+      curl localhost:8085/metrics
+      </pre>
```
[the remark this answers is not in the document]
mtabasko: I would suggest the following. Ideally, these steps should be numbered. Can "deploy/velero" be replaced by "{{cluster_name}}" here, so it matches the sample query on line 44?)

 To verify that {{app_name_short}} is exposing metrics on the expected endpoint, do the following:
1. Set up port forwarding by using the following command:
  <pre class="devsite-click-to-copy">
  kubectl -n {{namespace_name}} port-forward deploy/velero 8085
  </pre>
1. Access the endpoint \`localhost:8085/metrics\` by using the browser or the \`curl\` utility in another terminal session.

integrations/velero/documentation.yaml
```diff
@@ -0,0 +1,91 @@
+exporter_type: included
+app_name_short: Velero
+app_name: {{app_name_short}}
+app_site_name: Velero
+app_site_url: https://velero.io/
+exporter_name: the elero exporter
```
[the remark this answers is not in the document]
mtabasko: Based on line 8 in the "prometheus_metadata.yaml" file, should this be "the Velero Prometheus exporter"? (definitely need the capital "V" though! :)

integrations/velero/documentation.yaml
```diff
@@ -0,0 +1,91 @@
...
+    endpoints:
+  + - port: metrics
+      scheme: http
+      interval: 30s
+      path: /metrics
+    selector:
+      matchLabels:
+  +     deploy: velero
+additional_podmonitoring_info: |
+  Ensure that the port and matchLabels match that of the {{app_name_short}} pods you wish to monitor.
```
mtabasko, line 41: Ensure that the values of the \`port\` and \`matchLabels\` fields match those ....

mtabasko: want

integrations/velero/documentation.yaml
```diff
@@ -0,0 +1,91 @@
...
+  {{app_name_short}} installations include the `metrics` named port and label `deploy: velero` by default.
```
mtabasko, line 42: This is really hard to parse. I think it means something like this:

include a \`port\' value of \`metrics'` and a \`deploy\` label with the value \`velero\`

It might be clearer to say "By default, {{app_name_short}} installations include a \`port\' value of \`metrics'` and a \`deploy`\l abel with the value \`velero\`."

mtabasko (changes requested):

integrations/velero/documentation.yaml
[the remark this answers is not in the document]
yqlu: Text LGTM, but let's number this as per mary's suggestion

integrations/velero/documentation.yaml
yqlu: Hmmm, my only slight pushback to @mtabasko is that we typically say "X is listening on port 8085" or "X exposes the `metrics` port", never "X has a `port` value of `metrics`"

- "By default, {{app_name_short}} installations have a port named `metrics`"
- "By default, {{app_name_short}} installations expose metrics on a port named `metrics`"

sounds more natural. But I don't feel super strongly about it.

yqlu (approved):
LGTM after outstanding comments on formatting

integrations/velero/documentation.yaml
yqlu: OK I see that we're using this phrasing e.g. in the Jenkins PR too (https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/pull/425/commits/5c3e15f8b47c4bf5cdcd1c5daa2bf6253839bb70).

In that case, I don't mind keeping this as it is, and then if we do decide to tweak the wording, it's less churn to push something to all the affected configs at once as a follow-up

integrations/velero/documentation.yaml
```diff
@@ -0,0 +1,95 @@
...
+minimum_exporter_version: v1.3.0
+additional_prereq_info: |
+  {{app_name_short}} exposes Prometheus-format metrics automatically; you do not
+  have to install it separately. To verify that {{exporter_name}} is emitting
+  metrics on the expected endpoints, do the following:
+
+  1. Set up port forwarding by using the following command:
+
+  <pre class="devsite-click-to-copy">
+  kubectl -n velero port-forward {{cluster_name}} 8085
```
yqlu, line 18: I think this was right before? Typically cluster_name refers to the GKE cluster name, does it have a different meaning in the context of velero?

Or at least, when I installed velero to test these instructions locally, I was using `kubectl -n {{namespace_name}} port-forward deploy/velero 8085`

yqlu (approved):
Just one question, then LGTM

integrations/velero/documentation.yaml
algchoo: That's my mistake, should be good to go.

yqlu (approved):

PR #425, opened 2022-12-19
GMP Documentation, Alerts, and Dashboard filters: Jenkins

algchoo:
This PR includes the following for the Jenkins GMP integration:
* documentation
* alerts
* dashboard filters

### DIFF
--- a/dashboards/jenkins/jenkins-prometheus.json
+++ b/dashboards/jenkins/jenkins-prometheus.json
@@ -84,7 +84,7 @@
           "scorecard": {
             "blankView": {},
             "timeSeriesQuery": {
-              "prometheusQuery": "jenkins_executor_free_value{${Cluster},${Location},${Namespace}}"
+              "prometheusQuery": "sum(jenkins_executor_free_value)"
             }
           },
           "title": "Executor Free"
@@ -155,7 +155,7 @@
           "scorecard": {
             "blankView": {},
             "timeSeriesQuery": {
-              "prometheusQuery": "jenkins_executor_in_use_value{${Cluster},${Location},${Namespace}}"
+              "prometheusQuery": "sum(jenkins_executor_in_use_value)"
             }
           },
           "title": "Executor In Use"
@@ -167,20 +167,27 @@
       {
         "height": 2,
         "widget": {
-          "scorecard": {
-            "thresholds": [
+          "title": "Health Check",
+          "xyChart": {
+            "chartOptions": {
+              "mode": "COLOR"
+            },
+            "dataSets": [
               {
-                "color": "RED",
-                "direction": "BELOW",
-                "label": "",
-                "value": 1
+                "plotType": "LINE",
+                "targetAxis": "Y1",
+                "timeSeriesQuery": {
+                  "prometheusQuery": "jenkins_health_check_score"
+                }
               }
             ],
-            "timeSeriesQuery": {
-              "prometheusQuery": "jenkins_health_check_score{${Cluster},${Location},${Namespace}}"
+            "thresholds": [],
+            "timeshiftDuration": "0s",
+            "yAxis": {
+              "label": "y1Axis",
+              "scale": "LINEAR"
             }
-          },
-          "title": "Health Check"
+          }
         },
         "width": 3,
         "xPos": 0,
@@ -192,7 +199,7 @@
           "scorecard": {
             "blankView": {},
             "timeSeriesQuery": {
-              "prometheusQuery": "jenkins_plugins_failed{${Cluster},${Location},${Namespace}}"
+              "prometheusQuery": "sum(jenkins_plugins_failed)"
             }
           },
           "title": "Plugins Failed"
@@ -207,7 +214,7 @@
           "scorecard": {
             "blankView": {},
             "timeSeriesQuery": {
-              "prometheusQuery": "jenkins_node_offline_value{${Cluster},${Location},${Namespace}}"
+              "prometheusQuery": "sum(jenkins_node_offline_value)"
             }
           },
           "title": "Nodes Offline"
--- a/dashboards/jenkins/jenkins-prometheus.json
+++ b/dashboards/jenkins/jenkins-prometheus.json
@@ -1,6 +1,24 @@
 {
   "category": "CUSTOM",
+  "dashboardFilters": [
+    {
+      "filterType": "RESOURCE_LABEL",
+      "labelKey": "cluster",
+      "templateVariable": "Cluster"
+    },
+    {
+      "filterType": "RESOURCE_LABEL",
+      "labelKey": "location",
+      "templateVariable": "Location"
+    },
+    {
+      "filterType": "RESOURCE_LABEL",
+      "labelKey": "namespace",
+      "templateVariable": "Namespace"
+    }
+  ],
   "displayName": "Jenkins Prometheus Overview",
+  "labels": {},
   "mosaicLayout": {
     "columns": 12,
     "tiles": [
@@ -17,7 +35,7 @@
                 "plotType": "LINE",
                 "targetAxis": "Y1",
                 "timeSeriesQuery": {
-                  "prometheusQuery": "jenkins_job_building_duration"
+                  "prometheusQuery": "jenkins_job_building_duration{${Cluster},${Location},${Namespace}}"
                 }
               }
             ],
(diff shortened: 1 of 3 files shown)


integrations/jenkins/documentation.yaml
```diff
@@ -0,0 +1,77 @@
...
+  {{exporter_name}} exposes Prometheus-format metrics when the [metrics](https://plugins.jenkins.io/metrics/)
+  and [prometheus](https://plugins.jenkins.io/prometheus/) plugins are installed.
+  You can run the following checks to verify that {{app_name_short}} is exposing Prometheus-format metrics.
+
+   *  To determine if {{exporter_name}} is exposing metrics, run the following command:
+
+      <pre class="devsite-click-to-copy">
+      kubectl -n <namespace_name> port-forward <jenkins_pod_name> <jenkins_port>
+      curl localhost:<jenkins_port>/prometheus
+      </pre>
```
yqlu, line 19: OK, like the Kibana case, or our existing docs for [zookeeper](https://cloud.google.com/stackdriver/docs/managed-prometheus/exporters/zookeeper) / [consul](https://cloud.google.com/stackdriver/docs/managed-prometheus/exporters/consul), beyond saying "Jenkins exports Prometheus-format metrics when ... are installed, here's how you verify this", let's give a couple more hints of how to install these plugins in the first place.

Even though https://plugins.jenkins.io/prometheus has a "How to install" link, the 3 options there don't seem 100% applicable for jenkins running on a GKE cluster... I got confused myself when I tried to set this up and test these instructions, and came up with the following draft -- are there other methods you would recommend? 

---

If you installed [the Jenkins community helm chart](https://github.com/jenkinsci/helm-charts/tree/main/charts/jenkins), you may include the Prometheus plugin in the `controller.installPlugins` field. Alternatively, you may create your own {{app_name_short}} container image with the following Dockerfile:

```
FROM jenkins/jenkins:lts
RUN jenkins-plugin-cli --plugins prometheus metrics
```

integrations/jenkins/documentation.yaml
```diff
@@ -0,0 +1,77 @@
+exporter_type: included
+app_name_short: Jenkins
+app_name: {{app_name_short}}
+app_site_name: Jenkins
+app_site_url: https://www.jenkins.io/
+exporter_name: the Jenkins exporter
+exporter_pkg_name: jenkins
+exporter_repo_url: https://github.com/jenkinsci/prometheus-plugin
+additional_prereq_info: |
+  {{exporter_name}} exposes Prometheus-format metrics when the [metrics](https://plugins.jenkins.io/metrics/)
```
yqlu, line 10: Let's make this {{app_name_short}}.

We shouldn't use exporter_name at the start of the sentence because this will get expanded to "the Jenkins exporter" with uncapitalized the at the start of a sentence.

integrations/jenkins/documentation.yaml
```diff
@@ -0,0 +1,77 @@
...
+  and [prometheus](https://plugins.jenkins.io/prometheus/) plugins are installed.
+  You can run the following checks to verify that {{app_name_short}} is exposing Prometheus-format metrics.
+
+   *  To determine if {{exporter_name}} is exposing metrics, run the following command:
```
yqlu, line 14: Same comment as in https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/pull/423#discussion_r1063786149

integrations/jenkins/documentation.yaml
```diff
@@ -0,0 +1,77 @@
+exporter_type: included
+app_name_short: Jenkins
+app_name: {{app_name_short}}
+app_site_name: Jenkins
+app_site_url: https://www.jenkins.io/
+exporter_name: the Jenkins exporter
+exporter_pkg_name: jenkins
+exporter_repo_url: https://github.com/jenkinsci/prometheus-plugin
```
yqlu, line 8: We want to set minimum_exporter_version here if we're setting 2.0.11 in prometheus_metadata.yaml

integrations/jenkins/documentation.yaml
yqlu: Also if we are specifying 2.0.11 above, should we specify `prometheus:2.0.11` etc in the sample Dockerfile?

integrations/jenkins/documentation.yaml
```diff
@@ -0,0 +1,77 @@
...
+      app.kubernetes.io/part-of: google-cloud-managed-prometheus
+  spec:
+    endpoints:
+  + - port: 8080
+      scheme: http
+      interval: 30s
+      path: /prometheus
+    selector:
+      matchLabels:
+  +     app.kubernetes.io/name: jenkins
```
yqlu, line 39: Same comment as in https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/pull/422#discussion_r1062917614 -- No need for the + here and on line 33

And we should say something like [this](https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/blob/master/integrations/airflow/documentation.yaml#L42) to remind users to check this against the labels of their jenkins workload, which can vary depending on how they set it up

integrations/jenkins/documentation.yaml
```diff
@@ -0,0 +1,77 @@
...
+  apiVersion: monitoring.googleapis.com/v1
+  kind: PodMonitoring
+  metadata:
+    name: jenkins
+    labels:
+      app.kubernetes.io/name: jenkins
+      app.kubernetes.io/part-of: google-cloud-managed-prometheus
+  spec:
+    endpoints:
+  + - port: 8080
```
yqlu, line 33: We prefer named ports to numerical ports when possible ([context](https://github.com/GoogleCloudPlatform/prometheus-engine/releases/tag/v0.4.1)) -- is that possible in this case?

integrations/jenkins/prometheus_metadata.yaml
```diff
@@ -0,0 +1,60 @@
+platforms:
+  - type: GKE
+    launch_stage: GA
```
yqlu, line 3: This should be hidden until we publish the docs

dashboards/jenkins/jenkins-prometheus.json
```diff
@@ -137,7 +155,7 @@
           "scorecard": {
             "blankView": {},
             "timeSeriesQuery": {
-              "prometheusQuery": "jenkins_executor_in_use_value"
+              "prometheusQuery": "jenkins_executor_in_use_value{${Cluster},${Location},${Namespace}}"
```
yqlu, line 158: The scorecards and gauges here might run into the same issue as https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/pull/380#discussion_r1046173234

integrations/jenkins/documentation.yaml
yqlu: Let's also add `{:class=external}` to these two links

mtabasko: Concur!

integrations/jenkins/documentation.yaml
mtabasko: Concur!

mtabasko (commented):
Nothing really beyond what YQ already pointed out.

robinreneemac727:
This has nothing to do with me or my devices right?

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

dashboards/jenkins/jenkins-prometheus.json
```diff
@@ -84,7 +84,7 @@
           "scorecard": {
             "blankView": {},
             "timeSeriesQuery": {
-              "prometheusQuery": "jenkins_executor_free_value{${Cluster},${Location},${Namespace}}"
+              "prometheusQuery": "sum(jenkins_executor_free_value)"
```
yqlu, line 87: We want to have both the sum and the template variable, correct?

If someone filters down to a specific cluster within the GCP project, we want this to calculate `sum(jenkins_executor_free_value)` for all jenkins workloads within the cluster? Same goes for the other gauges in this commit

integrations/jenkins/documentation.yaml
yqlu: Sorry for the churn... same comment as Velero about numbering?

dashboards/jenkins/jenkins-prometheus.json
yqlu: Bump -- do the gauges need to be
- `sum(jenkins_executor_free_value{${Cluster},${Location},${Namespace}})`
- `sum(jenkins_executor_in_use_value{${Cluster},${Location},${Namespace}})`
- `sum(jenkins_plugins_failed{${Cluster},${Location},${Namespace}})`

for the filtering on the template variables to work?

yqlu (approved):

PR #427, opened 2022-12-19
GMP Documentation, Alerts, and Dashboard filters: etcd

algchoo:
This PR includes the following the etcd GMP integration:
* documentation
* alerts
* dashboard filters

### DIFF
--- a/integrations/etcd/documentation.yaml
+++ b/integrations/etcd/documentation.yaml
@@ -8,15 +8,17 @@ exporter_pkg_name: etcd
 exporter_repo_url: https://etcd.io/docs/v3.1/op-guide/monitoring/
 additional_prereq_info: |
   {{app_site_name}} exposes Prometheus-format metrics automatically; you do not
-  have to install it separately. You can run the following checks to verify
-  that etcd is exposing Prometheus-format metrics.
+  have to install it separately. To verify that {{exporter_name}} is emitting 
+  metrics on the expected endpoints, do the following:
 
-   *  To determine if {{app_site_name}} is exposing metrics, run the following command:
+  Set up port forwarding by using the following command:
 
-      <pre class="devsite-click-to-copy">
-      kubectl -n <namespace_name> port-forward <etcd_pod_name> 2379
-      curl localhost:2379/metrics
-      </pre>
+  <pre class="devsite-click-to-copy">
+  kubectl -n {{namespace_name}} port-forward {{pod_name}} 2379
+  </pre>
+
+  Access the endpoint `localhost:2379/metrics` by using the browser
+  or the `curl` utility in another terminal session.
 dashboard_available: true
 multiple_dashboards: false
 dashboard_display_name: {{app_name_short}} Prometheus Overview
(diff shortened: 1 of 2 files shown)


varun-c (approved):
Changes to etcd-prometheus.json and prometheus_metadata.yaml look good to me. Adding yqlu to review the documentation.yaml.

integrations/etcd/documentation.yaml
```diff
@@ -0,0 +1,79 @@
...
+      app.kubernetes.io/part-of: google-cloud-managed-prometheus
+  spec:
+    endpoints:
+  + - port: 2379
+      scheme: http
+      interval: 30s
+      path: /metrics
+    selector:
+      matchLabels:
+  +     app.kubernetes.io/name: etcd
```
yqlu, line 41: No "+"s needed in podmonitoring config because we're asking users to install one from scratch, not make changes to an existing setup.

yqlu: Additionally, it seems like there are many ways to deploy etcd which result in different labels.

For example, I followed this sample yaml - https://github.com/etcd-io/etcd/tree/main/hack/kubernetes-deploy - which used `app: etcd`, so we can't assume all etcd deployments will have `app.kubernetes.io/name: etcd`.

Let's add a paragraph to `additional_podmonitoring_info` reminding users to double-check that the labels match the selector, in the same style as [this](https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/blob/master/integrations/rabbitmq/documentation.yaml#L44), especially given that this is an `exporter_type: included` doc, and we are assuming they already have an etcd deployment up and running, and are not providing any sample yaml for it.

integrations/etcd/documentation.yaml
```diff
@@ -0,0 +1,83 @@
...
+      scheme: http
+      interval: 30s
+      path: /metrics
+    selector:
+      matchLabels:
+        app.kubernetes.io/name: etcd
+additional_podmonitoring_info: |
+  Ensure that the values of the `port` and `matchLabels` fields match those of the {{app_name_short}} pods you want to monitor.
+  Deploying {{app_name_short}} using this [example](https://github.com/etcd-io/etcd/tree/main/hack/kubernetes-deploy){:class=external}
+  will contain the label `app: etcd` and `client` port.
```
yqlu, line 45: One last thing - if this text says "Deploying etcd using this example will contain the label `app: etcd`", we should use `app: etcd` in the podmonitoring config (line 41 above)? Because this paragraph is explaining why the label selectors in the sample podmonitoring config we provide are what they are.

Or should we keep line 41 unchanged, and change this paragraph to match? This was just one example I found, but not sure if there are more common helm charts or installation methods out there that do label your pods with `app.kubernetes.io/name: etcd` by default?

algchoo: There is an existing helm chart packaged by Bitnami that will label the pods that way by default, I've updated the example to refer to that instead.

yqlu (approved):

PR #433, opened 2022-12-19
GMP Documentation, Alerts, and Dashboard filters: Apache Hadoop

algchoo:
This PR includes the following for the Hadoop GMP integration:
* documentation
* alerts
* dashboard filters

### DIFF
--- a/integrations/hadoop/documentation.yaml
+++ b/integrations/hadoop/documentation.yaml
@@ -1,5 +1,5 @@
 exporter_type: sidecar
-app_name_short: Apache Hadoop
+app_name_short: Hadoop
 app_name: Apache {{app_name_short}}
 app_site_name: Apache Hadoop
 app_site_url: https://hadoop.apache.org/
@@ -71,6 +71,9 @@ config_mods: |
   +         items:
   +         - key: config.yaml
   +           path: config.yaml
+additional_install_info: |
+  These instructions are based on changes made to a [helm chart](https://artifacthub.io/packages/helm/apache-hadoop-helm/hadoop).
+  The templates can be downloaded and altered. The example above is shown as if everything was in a single YAML file. 
 podmonitoring_config: |
   apiVersion: monitoring.googleapis.com/v1
   kind: PodMonitoring
--- a/integrations/hadoop/documentation.yaml
+++ b/integrations/hadoop/documentation.yaml
@@ -1,5 +1,5 @@
 exporter_type: sidecar
-app_name_short: Hadoop
+app_name_short: Apache Hadoop
 app_name: {{app_name_short}}
 app_site_name: Apache Hadoop
 app_site_url: https://hadoop.apache.org/
--- a/integrations/hadoop/documentation.yaml
+++ b/integrations/hadoop/documentation.yaml
@@ -1,7 +1,7 @@
 exporter_type: sidecar
 app_name_short: Hadoop
 app_name: {{app_name_short}}
-app_site_name: Hadoop
+app_site_name: Apache Hadoop
 app_site_url: https://hadoop.apache.org/
 exporter_name: the JMX exporter
 exporter_pkg_name: jmx-exporter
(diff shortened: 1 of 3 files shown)


integrations/hadoop/prometheus_metadata.yaml
```diff
@@ -0,0 +1,44 @@
+platforms:
```
varun-c, line 1: This file is missing the following metrics that are present in the [hadoop-prometheus-overview.json](https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/blob/40b6c1c95cffb0e9180db9180e870f7fcf848663/dashboards/hadoop/hadoop-prometheus-overview.json) dashboard:

- hadoop_namenode_missingblocks

Should that be added here?

varun-c (commented):
Reviewed the hadoop-prometheus-overview.json and prometheus_metadata.yaml files - they look good except for one missing metric in the metadata file.

Adding yqlu to review the documentation.yaml file.

integrations/hadoop/documentation.yaml
```diff
@@ -0,0 +1,143 @@
...
+      spec:
+        containers:
+        - name: hadoop-hdfs
+          image: "farberg/apache-hadoop:3.3.2"
+  +       env:
+  +         - name: HDFS_NAMENODE_OPTS
+  +           value: "-Dcom.sun.management.jmxremote=true -Dcom.sun.management.jmxremote.authenticate=false -Dcom.sun.management.jmxremote.ssl=false -Dcom.sun.management.jmxremote.port=1026"
+          command:
+          - "/bin/bash"
+          - "/tmp/hadoop-config/bootstrap.sh"
```
varun-c, line 46: When I tried following these instructions to set up the exporter for hadoop, I ran into an error because the hadoop configuration is incomplete. It looks like we need to define the bootstrap.sh script for hadoop and mount it as a hadoop-config volume.

We can either:
1. Add the complete config that will set up hadoop and then set up the exporter to export metrics
2. If the above is too much config and out of scope for the document, we should add a "..." to the hadoop-hdfs config and add an additional_install_info field similar to how we have for Redis, making it clear that we expect them to have a working hadoop application, or pointing to a tutorial if they don't.

integrations/hadoop/prometheus_metadata.yaml
```diff
@@ -0,0 +1,48 @@
...
+  - type: GKE
+    detections:
+    - characteristic_metric:
+        metric_type: prometheus.googleapis.com/hadoop_namenode_numlivedatanodes/unknown:counter
+    launch_stage: HIDDEN
+    exporter_metadata:
+      name: Hadoop Prometheus Exporter
```
varun-c, line 8:
```suggestion
      name: JMX Exporter
```

integrations/hadoop/documentation.yaml
algchoo: Apologies, the way Hadoop was deployed on our end was using this [helm chart](https://artifacthub.io/packages/helm/apache-hadoop-helm/hadoop). We deployed it, generated a YAML file, then updated that to include the exporter. 

I think we would want to specify that the helm chart was used and is necessary for the copy/pasting to work appropriately. If a user downloads the templates and updates them accordingly, that should work. 

I've added an `additional_install_info` field and made it clear that these instructions are based on modifying the templates from that helm chart.

varun-c: Got it - I was able to install it using the helm chart and complete the set up.

integrations/hadoop/documentation.yaml
```diff
@@ -0,0 +1,146 @@
...
+app_site_url: https://hadoop.apache.org/
+exporter_name: the JMX exporter
+exporter_pkg_name: jmx-exporter
+exporter_repo_url: https://github.com/bitnami/containers/tree/main/bitnami/jmx-exporter
+dashboard_available: true
+minimum_exporter_version: "0.17.0"
+multiple_dashboards: false
+dashboard_display_name: {{app_name_short}} Prometheus Overview
+config_mods: |
+  apiVersion: v1
```
varun-c, line 14: IIUC, lines 14-22 all need to be added to the existing config yaml. We should make that clear by appending a "+" to these lines.

varun-c (approved):

PR #437, opened 2022-12-19
GMP Documentation: Apache HTTP

algchoo:
This PR includes the following for the Apache HTTP GMP integration:
* documentation

### DIFF
--- a/integrations/apache/documentation.yaml
+++ b/integrations/apache/documentation.yaml
@@ -9,7 +9,7 @@ exporter_repo_url: https://github.com/Lusitaniae/apache_exporter
 dashboard_available: true
 minimum_exporter_version: v0.11.0
 multiple_dashboards: false
-dashboard_display_name: {{app_name_short}} Prometheus Overview
+dashboard_display_name: Apache Prometheus Overview
 config_mods: |
   apiVersion: v1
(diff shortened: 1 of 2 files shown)


integrations/apache/documentation.yaml
```diff
@@ -0,0 +1,91 @@
...
+    - port: prometheus
+      scheme: http
+      interval: 30s
+      path: /metrics
+    selector:
+      matchLabels:
+        app.kubernetes.io/name: httpd
+additional_prereq_info: |
+  Ensure that the values of the `port` and `matchLabels` fields match those of the {{app_name_short}} pods you want to monitor.
+alerts_config: {% includecode content_path="stackdriver/docs/managed-prometheus/exporters/configs/httpd_alerts.yaml" %}
```
varun-c, line 89: IIUC, this field refers to a separate yaml file that contains the alerts. Did you mean to add that file as part of this PR or will that be a separate PR? If it is the latter, we can remove this for now.

algchoo: The case here was that the existing alerts were not easily ported over using the available prometheus metrics. I've removed both sections. Thank you for the feedback, I appreciate it!

varun-c (approved):

PR #441, opened 2022-12-19
GMP Documentation, Alerts, and Dashboard filters: MongoDB

algchoo:
This PR includes the following for the MongoDB GMP integration: 
* documentation
* alerts
* dashboard filters


integrations/mongodb/documentation.yaml
```diff
@@ -0,0 +1,90 @@
...
+    labels:
+      app.kubernetes.io/component: rules
+      app.kubernetes.io/name: mongodb-rules
+      app.kubernetes.io/part-of: google-cloud-managed-prometheus
+  spec:
+    groups:
+    - name: mongodb
+      interval: 30s
+      rules:
+      - alert: MongoDBConnectionUtilizationNearMax
```
yqlu, line 79: I noticed there is a mongodb_up metric, can we add a very basic alert on mongodb is down? Similar to what we did for Redis (https://cloud.google.com/stackdriver/docs/managed-prometheus/exporters/redis#rules-alerts)

integrations/mongodb/documentation.yaml
```diff
@@ -0,0 +1,90 @@
...
+      interval: 30s
+      path: /metrics
+    selector:
+      matchLabels:
+        app.kubernetes.io/name: mongodb-exporter
+additional_prereq_info: |
+  The {{app_name_short}} exporter is configurable with [flags](https://hub.docker.com/r/percona/mongodb_exporter){:class=external},
+  which can be set as container args. You can adjust your exporter's configuration to suit your {{app_name_short}} configuration.
+
+  Update the `--mongodb.uri` flag to match your environment's username, password, and clusterIP service.
```
yqlu, line 63: This seems identical to additional_install_info? Let's omit lines 59 - 63 and just use additional_install_info. 

The primary difference between the twois that in the template this would render before standalone_config, while additional_install_info would render after standalone_config. In this case, the latter seems more appropriate.

yqlu (commented):
Thanks! I tested this against mongodb that I set up myself with the community helm chart, worked great.

integrations/mongodb/documentation.yaml
algchoo: I removed the `additional_prereq_info` section. I added a comment around the exporter not being compatible with MongoDB version 6.0.4. When I had that version running, the exporter only returned `mongodb_up`. I'm uncertain if there's something I'm missing where configuration is concerned.

I read this thread where it seems the exporter was able to work w/ version `5.0.6`, which is why I believe I may have misconfigured something. My steps, deploying using the community operator:
1. Updated the operator image tags and the version of MongoDB in the yaml
2. Left the `prometheus:` section out, so as to not setup and endpoint with auth
3. Setup the exporter the same way we had it setup for `4.2.6`

The result is that the exporter connects and returns some metrics, including `mongodb_up`, but there are no other metrics with a `mongodb` prefix.

yqlu (approved):

integrations/mongodb/documentation.yaml
yqlu: Thanks for flagging this! I've approved this for now so we can publish it and get value out of it. Let me look into the mongodb v6 issue on my end as well next week.

algchoo: I learned why I was only seeing the `mongodb_up` metric coming back from the exporter and that the community operator for MongoDB allows us to configure an endpoint that [exposes prometheus metrics](https://github.com/mongodb/mongodb-kubernetes-operator/blob/master/config/samples/mongodb.com_v1_mongodbcommunity_prometheus.yaml).

I misconfigured the exporter, initially, needed to pass the `--collect-all` argument, which was not needed prior.
```yaml
  args:
  - --mongodb.uri=mongodb://my-user:password@example-mongodb-svc:27017
  - --discovering-mode
  - --collect-all
```

After that, (testing v6.0.5 of MongoDB) the dashboard is working as expected. The endpoint that the operator allows us to configure forces us to require basic authentication. The PodMonitoring resource can't be setup with authentication, to my knowledge.

PR #442, opened 2022-12-19
Adds GKE Cluster scoped CPU/Memory/ContainerRestarts alerts

huygaa11:
Screenshot 2022-12-20 at 12 56 35 PM

Screenshot 2022-12-20 at 1 14 40 PM

### DIFF
--- a/alerts/google-gke/cpu-limit-utilization-containers-within-cluster.v1.json
+++ b/alerts/google-gke/cpu-limit-utilization-containers-within-cluster.v1.json
@@ -1,0 +1,30 @@
+{
+  "displayName": "GKE Container - High CPU Limit Utilization (${CLUSTER_NAME} cluster)",
+  "userLabels": {},
+  "conditions": [
+    {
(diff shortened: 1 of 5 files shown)


alerts/google-gke/cpu-limit-utilization-containers-within-cluster.v1.json
```diff
@@ -0,0 +1,30 @@
+{
+  "displayName": "GKE Container - High CPU Limit Utilization (${CLUSTER_NAME} cluster)",
```
huygaa11, line 2: @cocosheng , I really would like to use the template variable here, however, it makes the displayName in metadata.yaml and template different. Is that ok?

cocosheng: discussed offline that this is ok since we're deprecating the field in metadata.yaml

cocosheng (approved):

PR #443, opened 2022-12-20
Update observability alerts to reflect changes in metric definitions

fangyh20:
image
tested with screenshot.
alert created in pantheon also.


alerts/google-workload-manager/metadata.yaml
```diff
@@ -6,6 +6,6 @@ alert_policy_templates:
   version: 1
 -
   id: sap-hana-ha-availability-low
```
cocosheng, line 8: would it be better to update id to match the change?

fangyh20:
image
updated  SAP Hana high availability id

cocosheng (approved):

PR #447, opened 2023-01-03
Update redis-prometheus.json

EvanSimpson:
Adds separate chart for network input chart. It seems like charts only support a single prometheus query dataset, so separate charts appears to be the solution here.

image

### DIFF
--- a/dashboards/redis/redis-prometheus.json
+++ b/dashboards/redis/redis-prometheus.json
@@ -25,7 +25,7 @@
       {
         "height": 4,
         "widget": {
-          "title": "Network I/O",
+          "title": "Network I/O - Input",
           "xyChart": {
             "chartOptions": {
               "mode": "COLOR"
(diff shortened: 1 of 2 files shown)


dashboards/redis/redis-prometheus.json
```diff
@@ -35,7 +35,7 @@
                 "plotType": "LINE",
                 "targetAxis": "Y1",
                 "timeSeriesQuery": {
-                  "prometheusQuery": "rate(redis_net_output_bytes_total{${Cluster}, ${Location}, ${Namespace}} [5m])"
+                  "prometheusQuery": "rate(redis_net_input_bytes_total{${Cluster}, ${Location}, ${Namespace}} [5m])"
```
varun-c, line 38: We may want to add this to the corresponding prometheus_metadata.yaml file too as a default metric and also update the preview image.

varun-c (approved):

dashboards/redis/redis-prometheus.json
EvanSimpson: I added it to the metadata file. I'll see if I can find some time to stage a redis cluster to update the screenshot with actual data.

PR #481, opened 2023-02-27
Update GKE infrastructure dashboards for accuracy

yqlu:
1. GKE Compute Resources Cluster View

    - The "[CPU|Memory] [Request|Limit] Utilization" charts all suffered from the same issues, namely that they were trying to compute a ratio between a node-based numerator and a container-based denominator. These are not directly comparable because the memory consumed on a node > the memory consumed by all the containers within a node because the latter counts other system pieces. This led to confusion when for example Memory Limit Utilization reported a number above 100%. Changing the numerators to the k8s_container equivalent metrics makes this more comparable.

        Note that the "CPU Utilization" and "Memory Utilization" charts still compute a ratio where the numerator and denominator are both node-based; this has not changed.

    - The "Bandwidth" charts now group by top_level_controller_name instead of component_name. Upon investigation, the latter only existed occasionally on certain workloads (if a certain annotation was provided) while the latter is more consistent. Consequently, the charts are renamed from "Average container bandwidth" to "Average pod bandwidth" -- they capture the average bandwidth of pods within the workload represented by top_level_controller_name.

2. GKE Compute Resources Node View

    - kubernetes.io/node/memory/used_bytes reports [two numbers split up by memory_type = evictable | non-evictable](https://cloud.google.com/monitoring/api/metrics_kubernetes#kubernetes-kubernetes). This fixes a bug where the "Memory Util % per Node" and "Memory Usage per Node" charts mistakenly averaged evictable and non-evictable numbers instead of summing them up.

3. GKE Compute Resources Workload View

   - For consistency with the GKE Workloads graphs, the memory charts here only report the non-evictable memory usage (and the collapsible section name makes this explicit). However, we were doing it for just some charts but not others. Make this consistent everywhere.


dashboards/google-kubernetes-engine/gke-compute-resources-cluster-view.json
```diff
@@ -30,6 +30,7 @@
         "widget": {
```
johnbryan, line 30: Upon further thought I think all the charts in the first row need updates. There are two valid versions of this first chart:
- container cpu usage / node _allocatable_ cores (user space cpu for both numerator and denominator)
- node cpu usage / node _total_ cores (user+system cpu for numerator and denominator)

The current calculation could erroneously go >100% as I understand these metrics.

yqlu: Yes, I think [you're right](https://learnk8s.io/allocatable-resources).

I went for the second option to be consistent with what we report in the CPU / Memory charts of GKE OiC
Screenshot 2023-02-27 at 5 18 19 PM

johnbryan (approved):

pintofloyed (approved):

PR #485, opened 2023-03-06
Add MQL alert policy template sample

stephypy:


ddli-google:
The id entry in metadata.yaml should be "entries-exceeds" instead of "entries exceeds".

stephypy:
Thank you! just changed it

ddli-google (approved):

PR #503, opened 2023-03-16
PostgreSQL Prometheus Dashboard: Update metric names

algchoo:
Currently the metric names for the **Buffers** and **Checkpoints Requested/Scheduled** charts are missing the `_total` at the end

example:
`pg_stat_bgwriter_buffers_checkpoint` should be `pg_stat_bgwriter_buffers_checkpoint_total`

* updated metric names for Buffers chart
* updated metrics for the Checkpoints Requests/Scheduled

### DIFF
--- a/integrations/postgresql/prometheus_metadata.yaml
+++ b/integrations/postgresql/prometheus_metadata.yaml
@@ -37,24 +37,24 @@ platforms:
         prometheus_name: pg_stat_database_tup_deleted
         kind: CUMULATIVE
         value_type: DOUBLE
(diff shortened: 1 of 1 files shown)


varun-c (commented):
Can you also add these to new metrics to the corresponding prometheus metadata file (https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/blob/master/integrations/postgresql/prometheus_metadata.yaml)? We want the metrics used in the dashboards / alerts to also be mentioned in that file.

varun-c (approved):

PR #532, opened 2023-04-14
Update existing container restarts recommended alerts to a 5 minute window and threshold value of 1

stevezease:
The PR updates existing container restarts recommended alerts to a 5 minute window and threshold value of 1

### DIFF
--- a/alerts/google-gke/metadata.yaml
+++ b/alerts/google-gke/metadata.yaml
@@ -29,14 +29,14 @@ alert_policy_templates:
       platform: GCP
 -
   id: restarts-all-containers
-  description: "Monitors all containers in the current project clusters and alerts if a container restarts. (Restart count for the container is more than 0 within 1 minute window)"
+  description: "Monitors all containers in the current project clusters and alerts if a container restarts more than once. (Restart count for the container is more than 1 within 5 minute window)"
   version: 1
   related_integrations:
     - id: gke
       platform: GCP
 -
   id: restarts-containers-within-cluster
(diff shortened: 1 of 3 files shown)


cocosheng (approved):

stevezease:
I feel that it would be best to start with 5 minutes for the time being. 

While I can see the reasoning behind wanting to use 6m, a 6 minute window is not something I've found other sources using, while a 5 minute window is already a [fairly common practice among existing crashloopbackoff alerts on the web](https://www.google.com/search?q=%22rate%28kube_pod_container_status_restarts_total%5B5m%5D%29%22&ei=fGo9ZOyPHYmdptQP0s2GQA&ved=0ahUKEwjsveeQorH-AhWJjokEHdKmAQgQ4dUDCBA&uact=5&oq=%22rate%28kube_pod_container_status_restarts_total%5B5m%5D%29%22&gs_lcp=Cgxnd3Mtd2l6LXNlcnAQAzIFCAAQogQyCAgAEIkFEKIEMgUIABCiBDIFCAAQogQ6CggAEEcQ1gQQsANKBAhBGABQ6wVYjQdg-wdoAXABeACAAYUBiAHSAZIBAzEuMZgBAKABAcgBCMABAQ&sclient=gws-wiz-serp#ip=1)

I think deviating from the norm may raise more questions and so we would need to provide convincing evidence to justify doing so. 

The alternative is to go with 5m and then potentially make aware of the edge cases it doesn't capture.

Thinking over the edge case, from my understanding the policy would miss containers taking longer than 4 minutes and 40 seconds to crash as we would need 2 restarts in a 5 minute interval. 
(container restarts first time after waiting 10s, and then would need to crash within 4min40s to restart again given the 20s wait time).

PR #534, opened 2023-04-18
Update README for old anthos dashboard

baizhenyu:
Deprecation notice and redirect to new location


yqlu (approved):

yqlu:
On second thought, @EvanSimpson does this have implications on github README parsing?

EvanSimpson:
I think this should be fine - README parsing only looks at tables currently.

PR #542, opened 2023-05-01
Application Integration dashboards

ShubhamThakkar:


dashboards/google-application-integration/integration-executions.01.png
yqlu, line 1: General comment that applies to all the screenshots here -- is it possible to mock data such that all the charts in the screenshots have data populated? These are used as preview images, so if they're populated, it's easier to showcase the dashboard (e.g. which charts are line charts, stacked area charts, bar charts, what are the y axes, ...)

dashboards/google-application-integration/integration-executions.json
```diff
@@ -1,501 +1,417 @@
...
-                      }
-                    }
+        "width": 4,
+        "xPos": 8,
+        "yPos": 0
+      },
+      {
+        "height": 4,
+        "widget": {
+          "title": "Error (atleast one) containing executions per min",
```
yqlu, line 288: atleast -> at least

dashboards/google-application-integration/integration-executions.01.png
yqlu: This also serves as a verification that there are no typos or errors in the filter or query populating the chart

dashboards/google-application-integration/integration-executions.json
yqlu: Reading this again, I think the original phrasing was clearer? Executions with at least one error per min

dashboards/google-application-integration/integration-executions.01.png
ShubhamThakkar: Got it, thanks for the suggestion, have made the suggested changes.

One more thing - Is it possible to remove the process of attaching screenshots, it seems too lengthy. For every change, we have to take the screenshots again. Or somehow there should be a process to auto-generate these and show it to customer.

dashboards/google-application-integration/integration-executions.json
ShubhamThakkar: Changed it.

dashboards/google-application-integration/data-processed.json
```diff
@@ -0,0 +1,220 @@
...
+    "dashboardFilters": [
+      {
+        "filterType": "RESOURCE_LABEL",
+        "labelKey": "location",
+        "templateVariable": ""
+      },
+      {
+        "filterType": "RESOURCE_LABEL",
+        "labelKey": "integration",
+        "stringValue": "shubham-monitoring-integration",
```
yqlu, line 12: Omit this line

yqlu (approved):
Just one small change, then LGTM

dashboards/google-application-integration/integration-executions.01.png
yqlu: Unfortunately, screenshots are an integral part of how sample and integration dashboards are presented in the UI. We are trying to keep the churn to a minimum by only mandating screenshots be provided for the first submission of each dashboard and when there are major layout changes, and not when you're just making minor edits to an existing dashboard. But it's hard to auto-generate them on the backend when (as we discussed) you need to be running a workload to generate values for some of these metrics.

dashboards/google-application-integration/data-processed.json
ShubhamThakkar: Removed it, thanks for catching it.

Thanks for reviewing the CL.

PR #548, opened 2023-05-09
Update metadata files for cloudsql and anthos config management

pintofloyed:


dashboards/google-cloudsql/metadata.yaml
```diff
@@ -2,7 +2,7 @@ sample_dashboards:
   -
     category: Cloud SQL
     id: cloudsql-general
-    display_name: Cloud SQL General
+    display_name: CloudSQL - General
```
varun-c, line 5: Can you update these names in the readme files too? We can attempt to keep them in sync for now.

varun-c (approved):

dashboards/google-cloudsql/metadata.yaml
pintofloyed: Done.

PR #549, opened 2023-05-11
Add Anthos apiserver latency graphs

leonzz:
add 4 graphs for apiserver latencies:
- API Server Webhook Latency [MEDIAN]
- API Server Webhook Latency [95th Percentile]
- API Server Request Latency [MEDIAN]
- API Server Request Latency [95th Percentile]

All 4 graphs are grouped by project, location, cluster and namespace.    namespace is used to distinguish Anthos on VMware user clusters with control plane V1 (kubeception).


leonzz:
cc @baizhenyu @logicalhan

logicalhan:
Can we also add alerts for webhook latencies which exceed 2s for over 5 minutes?

leonzz:
> Can we also add alerts for webhook latencies which exceed 2s for over 5 minutes?
We won't be able to create alerts here, since this repo is all about dashboards. We can discuss more about where & how the alerts would work in a separate effort.

CatherineF-dev:
fyi: there is a new feature named "Alert Chart" which can add alerts into a dashboard.  But we haven't used this chart in our dashboards before. We can consider it later since need to consider setting up notification channels.
```
  "widget": {
          "alertChart": {
            "name": ""
          }
        },
```

baizhenyu (approved):

baizhenyu (approved):

yqlu (approved):
LGTM

Optional follow-up: consider if we can improve info density by unioning median and 95th percentile as different series in the same chart? Not sure if this is complicated because each chart is already the union of the k8s_container and prometheus_target versions of the same metric.

PR #550, opened 2023-05-12
updating desc text and adding limit template

google-nalin:
Adding UX approved text and a new limit based alert template
image

### DIFF
--- a/alerts/google-quotas/metadata.yaml
+++ b/alerts/google-quotas/metadata.yaml
@@ -24,16 +24,19 @@ alert_policy_templates:
   description: "This policy will notify you if quota usage reaches the alert trigger threshold of 80%. Expand the drop-down below if you would like to edit the alert trigger threshold or further customize this alert policy."
   version: 1
 -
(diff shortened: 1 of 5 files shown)


alerts/google-quotas/metadata.yaml
```diff
@@ -24,16 +24,19 @@ alert_policy_templates:
   description: "This policy will notify you if quota usage reaches the alert trigger threshold of 80%. Expand the drop-down below if you would like to edit the alert trigger threshold or further customize this alert policy."
   version: 1
 -
-  id: all-quota-adjuster-changes
-  description: "This policy will notify you of any changes made by Quota Adjuster"
+  id: all-adjustements-by-quota-adjuster
```
cocosheng, line 27: are the json file names updated to match these changes too?

google-nalin: Yes the json file names are updated to match these changes

cocosheng (approved):

PR #579, opened 2023-07-06
Updated the Apache Flink documentation

algchoo:
**Changes**
* [updated the flink documentation to use the latest configuration for the Prometheus reporter](https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/commit/5726528cadc31a25ebd63d65ddc0d3b977d8cb2c)

**Details**

For Apache Flink, version `1.17.x` the appropriate way to configure prometheus metrics is to set the following in the configuration file:
```
metrics.reporter.prom.factory.class: org.apache.flink.metrics.prometheus.PrometheusReporterFactory
```
Previously, it was this
```
metrics.reporter.prom.class: org.apache.flink.metrics.prometheus.PrometheusReporter
```

### DIFF
--- a/integrations/flink/documentation.yaml
+++ b/integrations/flink/documentation.yaml
@@ -9,7 +9,7 @@ minimum_exporter_version: "1.15"
 exporter_repo_url: https://flink.apache.org/features/2019/03/11/prometheus-monitoring.html
 additional_prereq_info: |
   {{app_name_short}} exposes Prometheus-format metrics when configured with
-  `metrics.reporter.prom.class: org.apache.flink.metrics.prometheus.PrometheusReporter`.
+  `metrics.reporter.prom.factory.class: org.apache.flink.metrics.prometheus.PrometheusReporterFactory`.
   If you deployed Flink with the official [getting started manifests](https://nightlies.apache.org/flink/flink-docs-master/docs/deployment/resource-providers/standalone/kubernetes/){:class="external"},
   add this new option to the ConfigMap:
   <pre>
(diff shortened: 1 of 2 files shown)


integrations/flink/documentation.yaml
```diff
@@ -36,7 +36,7 @@ additional_prereq_info: |
     flinkVersion: v1_15
```
EvanSimpson, line 36: Should we update all of the references to flink 1.15 to be 1.17 as well (if this change isn't backwards compatible)? If so, then let's also update it in the prometheus metadata too https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/blob/master/integrations/flink/prometheus_metadata.yaml#L10

EvanSimpson (changes requested):

yqlu:
FYI @mtabasko for when this gets submitted, since the pipeline for exporter doc publishing isn't fully automated

integrations/flink/documentation.yaml
algchoo: Changes pushed, thanks for catching that!

EvanSimpson: Doubling checking - should the image and flinkVersion values here be updated as well?

algchoo: Yes, it should, thank you!

EvanSimpson: Great, thanks.

EvanSimpson (approved):

mtabasko:
Send a CL to Evan and YQ for the docs update!

PR #580, opened 2023-07-06
Updated Apache HTTP documentation

algchoo:
**Changes**
* [removed args for the exporter due to them all being default values, updated min version](https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/commit/d0f0618581428d80aad07a55fa54edbf0fdae3c3) 
* [updated the args to include web.listen-address which replaced telemetry.address](https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/commit/960e5356b6b58b0469073cbe7945981d71943072) 

**Details**
In the latest version of the exporter, the `--telemetry.address` flag is replaced by `--web.listen-address`. Wasn't initially clear to me from the docker container documentation and the github documentation was more helpful. I thought it initially had something to do with the flags having a certain number of dashes and I noticed we we're using default values, so I just removed the args, but that wasn't the issue.

Another note, I didn't see the dashboard appear in the **Integrations** tab, as I believe we would expect. I was able to verify that metrics were flowing using a clone of the dashboard.

### DIFF
--- a/integrations/apache/documentation.yaml
+++ b/integrations/apache/documentation.yaml
@@ -7,7 +7,7 @@ exporter_name: the Apache HTTP exporter
 exporter_pkg_name: httpd_exporter
 exporter_repo_url: https://github.com/Lusitaniae/apache_exporter
 dashboard_available: true
-minimum_exporter_version: v0.11.0
+minimum_exporter_version: v1.0.0
 multiple_dashboards: false
 dashboard_display_name: Apache Prometheus Overview
 config_mods: |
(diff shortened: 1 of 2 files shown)


yqlu (approved):
I tested the new config in my toy cluster and verified that scraping metrics worked OOTB. Thanks for making this change!

> Another note, I didn't see the dashboard appear in the Integrations tab, as I believe we would expect.

Good catch! Looks like that's because this integration is still marked as "HIDDEN". Can you change line 3 of `prometheus_metadata.yaml` from HIDDEN to GA? Seems like all the other prometheus integrations are correctly marked already, and only this one fell between the cracks.

PR #581, opened 2023-07-07
Updated the PostgreSQL Prometheus documentation and dashboard

algchoo:
**Changes**
* [updated postgresql dashboard json to account for changes in the exporter](https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/commit/6bb1be830942eefe8c6f73eb33d34fb9ca5f01ce)
* [updated docs and metadata to account for changes in the exporter](https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/commit/5908985ff3bed385932235a8785b02525cecc485)

**Details**
> As a result of downloading the json, editing that, and re-uploading, a few extra keys are now present, unsure if removing them is a good idea (ie. breakdowns, dimensions, measures/etc)

During testing, we learned that 
* the postgres dashboard json had errors, on top of the changes that were needed as a result of changes in the exporter. It was all just incorrect metrics in various charts, or duplicates, should be good to go now
* Certain metrics coming from the exporter are now [enabled/disabled with a flag](https://github.com/prometheus-community/postgres_exporter#flags) and so the flag was added as an argument in the documentation.
* Various metrics had a `_total` suffix added to them and updates were made to the dashboard json to account for this.

### DIFF
--- a/integrations/postgresql/prometheus_metadata.yaml
+++ b/integrations/postgresql/prometheus_metadata.yaml
@@ -7,7 +7,7 @@ platforms:
     exporter_metadata:
       name: GKE Prometheus Exporter
       doc_url: https://github.com/prometheus-community/postgres_exporter
-      minimum_supported_version: v0.14.0
+      minimum_supported_version: v0.11.0
     default_metrics:
       - name: prometheus.googleapis.com/pg_stat_database_numbackends/gauge
         prometheus_name: pg_stat_database_numbackends
--- a/integrations/postgresql/prometheus_metadata.yaml
+++ b/integrations/postgresql/prometheus_metadata.yaml
@@ -7,7 +7,7 @@ platforms:
     exporter_metadata:
       name: GKE Prometheus Exporter
       doc_url: https://github.com/prometheus-community/postgres_exporter
-      minimum_supported_version: v0.13.1
+      minimum_supported_version: v0.14.0
     default_metrics:
       - name: prometheus.googleapis.com/pg_stat_database_numbackends/gauge
         prometheus_name: pg_stat_database_numbackends
--- a/integrations/postgresql/prometheus_metadata.yaml
+++ b/integrations/postgresql/prometheus_metadata.yaml
@@ -7,7 +7,7 @@ platforms:
     exporter_metadata:
       name: GKE Prometheus Exporter
       doc_url: https://github.com/prometheus-community/postgres_exporter
-      minimum_supported_version: v0.11.1
+      minimum_supported_version: v0.13.1
     default_metrics:
       - name: prometheus.googleapis.com/pg_stat_database_numbackends/gauge
         prometheus_name: pg_stat_database_numbackends
--- a/dashboards/postgresql/postgresql-prometheus.json
+++ b/dashboards/postgresql/postgresql-prometheus.json
@@ -17,7 +17,7 @@
       "templateVariable": "Namespace"
     }
   ],
-  "displayName": "PostgreSQL Prometheus Overview",
+  "displayName": "PostgreSQL Prometheus Overview - Q2 2023 Currency Testing Fix",
   "labels": {},
   "mosaicLayout": {
     "columns": 12,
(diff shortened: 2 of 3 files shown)


integrations/postgresql/prometheus_metadata.yaml
```diff
@@ -7,7 +7,7 @@ platforms:
     exporter_metadata:
       name: GKE Prometheus Exporter
       doc_url: https://github.com/prometheus-community/postgres_exporter
-      minimum_supported_version: v0.11.1
+      minimum_supported_version: v0.13.1
     default_metrics:
       - name: prometheus.googleapis.com/pg_stat_database_numbackends/gauge
         prometheus_name: pg_stat_database_numbackends
```
yqlu, line 13: Note that the list of metrics below need to be updated as well

yqlu (commented):
@EvanSimpson let's discuss this next week -- of the current wave of exporter doc updates this one involves changing the metric names outright, thereby breaking the dashboards. 

It feels wrong to bump the db version, exporter version, update the dashboard to a new working state but everyone who set up the older version of the integration against postgres 14 now has a non-functional dashboard (especially since the integration detection using the characteristic metric pg_settings_max_connections will continue to work against both old and new setups).

yqlu:
Thanks for investigating! Could this be related to https://github.com/prometheus-community/postgres_exporter/issues/818 ? From the comments on that thread, that issue can be spotted via the logs on the exporter.

If so, maybe we should hold off on updating the documentation to support 0.13.x for now?

Sticksman:
let me add a logger to it, maybe it'll spit out some errors

algchoo:
@yqlu I think that for now we should hold off, as you suggested. Would you like this PR closed for now? I left a comment on [this thread](https://github.com/prometheus-community/postgres_exporter/pull/685#issuecomment-1652326684) regarding what I'm seeing for metrics when using PostgreSQL 15.3 and version `0.13.2` of the exporter. I was still unable to see the expected `pg_stat_database` metrics.

algchoo:
@yqlu It was determined that there is a race condition that causes some collectors to fail in the exporter. The latest information is [here](https://github.com/prometheus-community/postgres_exporter/pull/685#issuecomment-1658887068). I went ahead and tested out the exporter after letting PostgreSQL start completely and after that, I saw the expected metrics. That means the dashboard doesn't have to be updated and with a fix for the problem on the way I am wondering how we'd like to handle that in the docs, if at all?

yqlu:
As discussed, let's:
- leave `minimum_supported_version` untouched in both documentation.yaml and prometheus_metadata.yaml
- if the metrics are not going to change, leave the dashboard untouched
- hold off on the version bumps in the recommended yaml snippet, but keep an eye on that thread... once we have an operational example (maybe not v0.13.2, but [a future version that includes this fix](https://github.com/prometheus-community/postgres_exporter/issues/880)?) then update [the yaml snippet](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/main/examples/postgresql/exporter.yaml.snippet) hosted on the GMP repo directly -- the official docs point to that copy.

Sticksman:
Watch for [this](https://github.com/prometheus-community/postgres_exporter/pull/882) to be merged and released

algchoo:
> As discussed, let's:
> 
> * leave `minimum_supported_version` untouched in both documentation.yaml and prometheus_metadata.yaml
> * if the metrics are not going to change, leave the dashboard untouched
> * hold off on the version bumps in the recommended yaml snippet, but keep an eye on that thread... once we have an operational example (maybe not v0.13.2, but [a future version that includes this fix](https://github.com/prometheus-community/postgres_exporter/issues/880)?) then update [the yaml snippet](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/main/examples/postgresql/exporter.yaml.snippet) hosted on the GMP repo directly -- the official docs point to that copy.

The PR that was mentioned has been merged, but the tag for the container that includes the fix is only on `master` and not a specific version. I'll continue to keep an eye out for the next release and will update the yaml snippet when that happens.

algchoo:
@yqlu The release of `0.14.0` has happened. Here's the [PR](https://github.com/GoogleCloudPlatform/prometheus-engine/pull/574) for updating the yaml snippet. Did we want to also get this PR updated/merged?

yqlu (approved):

PR #584, opened 2023-07-12
Updated Consul Prometheus documentation

algchoo:
**Changes**
* [updated consul image due to name change, updated minimum version](https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/commit/236e6b403ebf8dc49a6766092b23f4ba6cd0987e)
* [updated the image name in the documentation for consul](https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/commit/9a622c7487d930db97c2bb996ff75234d693f89e)

**Details**
Updated the minimum version for the exporter in the documentation and metadata. Due to the image name changing, that is updated (just noticed that I missed this part, sorry for the confusing commits) in the documentation, as well.

### DIFF
--- a/integrations/consul/documentation.yaml
+++ b/integrations/consul/documentation.yaml
@@ -7,7 +7,7 @@ exporter_name: the Consul exporter
 exporter_pkg_name: consul
 exporter_repo_url: https://developer.hashicorp.com/consul/docs/agent/config/config-files#telemetry-parameters
 dashboard_available: true
-minimum_exporter_version: "1.13.3"
+minimum_exporter_version: "1.16.0"
 multiple_dashboards: false
 dashboard_display_name: {{app_name_short}} Prometheus Overview
 config_mods: |
@@ -38,7 +38,7 @@ config_mods: |
       spec:
         containers:
         - name: consul
-          image: consul:1.13.3
+          image: consul:1.16.0
           ports:
           - containerPort: 8500
             name: consul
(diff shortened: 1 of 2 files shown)


yqlu:
Just double-checking, why is this version bump necessary? Back when we published this, we did verify that consul 1.13.x does work and you can get the necessary metrics scraped by GMP?

Should we only bump minimum supported exporter versions when there is actually a breaking change in the config (like in #580) or if the new version emits some new metrics that we are showcasing in the dashboard?

algchoo:
> Just double-checking, why is this version bump necessary? Back when we published this, we did verify that consul 1.13.x does work and you can get the necessary metrics scraped by GMP?
> 
> Should we only bump minimum supported exporter versions when there is actually a breaking change in the config (like in #580) or if the new version emits some new metrics that we are showcasing in the dashboard?

I agree with you on this, but wasn't sure initially. These changes do not have an effect on the dashboard/alerts, no other breaking change that I'm aware of exists. I'll update this.

algchoo:
> > Just double-checking, why is this version bump necessary? Back when we published this, we did verify that consul 1.13.x does work and you can get the necessary metrics scraped by GMP?
> > Should we only bump minimum supported exporter versions when there is actually a breaking change in the config (like in #580) or if the new version emits some new metrics that we are showcasing in the dashboard?
> 
> I agree with you on this, but wasn't sure initially. These changes do not have an effect on the dashboard/alerts, no other breaking change that I'm aware of exists. I'll update this.

If we are keeping the minimum version as is, I don't see a reason to change the documentation either, even with the newer image having a slightly different name, I'll close this PR.

PR #596, opened 2023-07-26
Updated MySQL Prometheus documentation

algchoo:
## **Changes**
* [updated the minimum exporter version, example configuration, and additional install info](https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/commit/81a4c3d5d2bff3f78ef547fa3ae91868c9f5b733) 
* [updated the minimum supported version in metadata](https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/commit/361e1136fdd66b71cde0a5ae1bc0cb9772f2a0f7)

## **Details**
The latest version of the MySQL Prometheus exporter works with MySQL 5.7 and MySQL 8.0.34. This new version of the exporter removes the `DATA_SOURCE_NAME` environment variable and instead we can pass in the `--mysqld.username` argument and the `MYSQLD_EXPORTER_PASSWORD` environment variable for authentication, or we can mount a configuration file with the credentials in it and pass the `--config.my-cnf=<PATH_TO_MOUNTED_CONFIG_FILE>`. 

Example configuration file:
```
[client]
user=root
password=password
```

I'm not sure which option to go with for the example configuration and wanted to get an opinion, we could change it to look something like
```
config_mods: |
apiVersion: v1
kind: ConfigMap
metadata:
  name: mysql-exporter-config
data:
  .my-cnf: |
    [client]
    user=root
    password=password
-------
  apiVersion: apps/v1
  kind: StatefulSet
  metadata:
    name: mysql
  spec:
    serviceName: mysql
    selector:
      matchLabels:
  +     app.kubernetes.io/name: mysql
    template:
      metadata:
        labels:
  +       app.kubernetes.io/name: mysql
      spec:
        containers:
  +     - name: exporter
  +       image: prom/mysqld-exporter:v0.15.0
  +       args:
  +         - --config.my-cnf=<PATH_TO_MOUNTED_CONFIGURATION>
  +       ports:
  +       - containerPort: 9104
  +         name: prometheus
  +       volumeMounts:
  +       - mountPath: /home/.my.cnf
  +         subPath: .my.cnf
  +         name: mysql-exporter-config
        - image: mysql:5.7
          name: mysql
          env:
          - name: MYSQL_ROOT_PASSWORD
            value: password
          - name: MYSQL_USER
            value: sbtest
          - name: MYSQL_PASSWORD
            value: password
          - name: MYSQL_DATABASE
            value: sbtest
          ports:
          - containerPort: 3306
            name: mysql
```

### DIFF
--- a/integrations/mysql/documentation.yaml
+++ b/integrations/mysql/documentation.yaml
@@ -20,7 +20,7 @@ config_mods: |
   +     [client]
   +     user=root
   +     password=password
-  -------
+  ---
   apiVersion: apps/v1
   kind: StatefulSet
   metadata:
@@ -48,6 +48,7 @@ config_mods: |
   +         subPath: my.cnf
   +         name: mysql-exporter-config
         - name: mysql
+          image: mysql:5.7
           env:
           - name: MYSQL_ROOT_PASSWORD
             value: password
@@ -61,9 +62,11 @@ config_mods: |
           - containerPort: 3306
   +     volumes:
   +     - name: mysql-exporter-config
-  +       items:
-  +       - key: my.cnf
-  +         path: my.cnf
+  +       configMap:
+  +         name: mysql-exporter-config
+  +         items:
+  +         - key: my.cnf
+  +           path: my.cnf
 podmonitoring_config: |
   apiVersion: monitoring.googleapis.com/v1
   kind: PodMonitoring
--- a/integrations/mysql/documentation.yaml
+++ b/integrations/mysql/documentation.yaml
@@ -7,7 +7,7 @@ exporter_name: the MySQL exporter
 exporter_pkg_name: mysqld_exporter
 exporter_repo_url: https://github.com/prometheus/mysqld_exporter
 dashboard_available: true
-minimum_exporter_version: v0.14.0
+minimum_exporter_version: v0.15.0
 multiple_dashboards: false
 dashboard_display_name: {{app_name_short}} Prometheus Overview
 config_mods: |
@@ -27,10 +27,12 @@ config_mods: |
       spec:
         containers:
   +     - name: exporter
-  +       image: prom/mysqld-exporter:v0.14.0
+  +       image: prom/mysqld-exporter:v0.15.0
+  +       args:
+  +         - --mysqld.username=root
   +       env:
-  +         - name: DATA_SOURCE_NAME
-  +           value: root:password@(localhost:3306)/
+  +         - name: MYSQLD_EXPORTER_PASSWORD
+  +           value: password
   +       ports:
   +       - containerPort: 9104
   +         name: prometheus
@@ -68,7 +70,7 @@ podmonitoring_config: |
 additional_prereq_info: |
   For information about creating a least-privileged user, see [Required Grants](https://github.com/prometheus/mysqld_exporter#required-grants){:class=external}.
 additional_install_info:
-  Update the `DATA_SOURCE_NAME` environment variable with credentials that work with your {{app_name}} instance.
+  Update the `--mysqld.username` argument and the `MYSQLD_EXPORTER_PASSWORD` environment variable with credentials that work with your {{app_name}} instance 
 sample_promql_query: up{job="mysql", cluster="{{cluster_name}}", namespace="{{namespace_name}}"}
 alerts_config: |
   apiVersion: monitoring.googleapis.com/v1
--- a/integrations/mysql/documentation.yaml
+++ b/integrations/mysql/documentation.yaml
@@ -11,45 +11,42 @@ minimum_exporter_version: v0.15.0
 multiple_dashboards: false
 dashboard_display_name: {{app_name_short}} Prometheus Overview
 config_mods: |
-  apiVersion: apps/v1
-  kind: StatefulSet
+  apiVersion: v1
+  kind: ConfigMap
   metadata:
-    name: mysql
-  spec:
-    serviceName: mysql
-    selector:
-      matchLabels:
-  +     app.kubernetes.io/name: mysql
-    template:
-      metadata:
-        labels:
-  +       app.kubernetes.io/name: mysql
-      spec:
-        containers:
-  +     - name: exporter
-  +       image: prom/mysqld-exporter:v0.15.0
-  +       args:
-  +         - --mysqld.username=root
-  +       env:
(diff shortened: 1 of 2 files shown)


yqlu:
Looking at the mysql-exporter README, I think the config map approach is more recommended as a best practice for security:

> To avoid putting sensitive information like username and password in the URL, you can have multiple configurations in config.my-cnf file and match it by adding `&auth_module=<section>` to the request.

On the topic of best practices, I see that we're defining a non-root user for mysql in this toy example:

```
          - name: MYSQL_USER
            value: sbtest
          - name: MYSQL_PASSWORD
            value: password
```

So should we use that instead of root for the exporter's auth?

Happy to help test this if you update the yaml!

algchoo:
> Looking at the mysql-exporter README, I think the config map approach is more recommended as a best practice for security:
> 
> > To avoid putting sensitive information like username and password in the URL, you can have multiple configurations in config.my-cnf file and match it by adding `&auth_module=<section>` to the request.
> 
> On the topic of best practices, I see that we're defining a non-root user for mysql in this toy example:
> 
> ```
>           - name: MYSQL_USER
>             value: sbtest
>           - name: MYSQL_PASSWORD
>             value: password
> ```
> 
> So should we use that instead of root for the exporter's auth?
> 
> Happy to help test this if you update the yaml!

Sounds good, I will update the documentation to show the example that uses the configuration file. The non-root user that is defined is used while running `sysbench` to generate load. I attempted to use that user for to authenticate against the exporter and they didn't have the appropriate permissions, so I went back to using the `root` user.

algchoo:
@yqlu For this [repo](https://github.com/observIQ/google-prometheus-environment) name of the branch that'll have changes is `dashboards/currency-test`.

integrations/mysql/documentation.yaml
```diff
@@ -7,47 +7,46 @@ exporter_name: the MySQL exporter
 exporter_pkg_name: mysqld_exporter
 exporter_repo_url: https://github.com/prometheus/mysqld_exporter
 dashboard_available: true
-minimum_exporter_version: v0.14.0
+minimum_exporter_version: v0.15.0
 multiple_dashboards: false
 dashboard_display_name: {{app_name_short}} Prometheus Overview
 config_mods: |
-  apiVersion: apps/v1
```
yqlu, line 14: As per the other example, we can update https://github.com/GoogleCloudPlatform/prometheus-engine/blob/main/examples/mysql/exporter.yaml.snippet directly (after addressing the other comments here)

But unlike the consul example, this PR doesn't become completely obsolete -- we do still need to update this file because of the change you made in line 70 (additional_install_info).

integrations/mysql/documentation.yaml
```diff
@@ -7,47 +7,46 @@ exporter_name: the MySQL exporter
...
+    +       image: prom/mysqld-exporter:v0.15.0
+    +       args:
+    +         - --config.my-cnf=/home/my.cnf
+    +       ports:
+    +       - containerPort: 9104
+    +         name: prometheus
+    +       volumeMounts:
+    +       - mountPath: /home/my.cnf
+    +         subPath: my.cnf
+    +         name: mysql-exporter-config
```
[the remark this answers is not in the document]
algchoo: I've updated the yaml to include a `volumeMount` and I'm wondering, do we want the `+` symbols with the `configmap` that is included in the example as well?

integrations/mysql/documentation.yaml
algchoo: Sounds good, just for clarity I wanted to make sure that the line you're mentioning here doesn't need to be changed? We just want to update things in this repo and the other one, right?

yqlu: Yep

integrations/mysql/documentation.yaml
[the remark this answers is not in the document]
yqlu: I think so.

In some cases like [the nginx one](https://cloud.google.com/stackdriver/docs/managed-prometheus/exporters/nginx), you can reasonably expect that anyone setting up nginx on k8s will have a configmap set up, so we just use "+"s to indicate the lines they need to add to the config map.

In this case, if you already had a mysql instance up and running, and then you were adding an exporter, you wouldn't already have a configmap with credentials beforehand, so the volume, volumemount, configmap are all indicated with "+"s.

Does that make sense?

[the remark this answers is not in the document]
algchoo: I believe so, I pushed changes to include the `+` symbols with the `configmap` unsure if we need all of them.

integrations/mysql/documentation.yaml
```diff
@@ -11,59 +11,59 @@ minimum_exporter_version: v0.14.0
...
+  + apiVersion: v1
+  + kind: ConfigMap
+  + metadata:
+  +  name: mysql-exporter-config
+  + data:
+  +   my.cnf: |
+  +     [client]
+  +     user=root
+  +     password=password
   -------
```
yqlu, line 23: Three dashes only please

integrations/mysql/documentation.yaml
```diff
@@ -47,7 +59,11 @@ config_mods: |
             value: sbtest
           ports:
           - containerPort: 3306
-            name: mysql
+  +     volumes:
+  +     - name: mysql-exporter-config
+  +       items:
+  +       - key: my.cnf
+  +         path: my.cnf
```
yqlu, line 66:
```suggestion
  +     volumes:
  +     - name: mysql-exporter-config
  +       configMap:
  +         name: mysql-exporter-config
  +         items:
  +         - key: my.cnf
  +           path: my.cnf
```

integrations/mysql/documentation.yaml
```diff
@@ -27,15 +37,17 @@ config_mods: |
...
   +       ports:
   +       - containerPort: 9104
   +         name: prometheus
-        - image: mysql:5.7
-          name: mysql
+  +       volumeMounts:
+  +       - mountPath: /home/my.cnf
+  +         subPath: my.cnf
+  +         name: mysql-exporter-config
+        - name: mysql
```
yqlu, line 50: Let's include `        - image: mysql:5.7` here

This is presented as a diff, so it's not super crucial and not meant to be operational JSON, but still good to document which sql version we tested against

yqlu (approved):

PR #635, opened 2023-10-06
Fix no data issue with the nvidia dashboard

LujieDuan:
b/296253311

For `GPU Memory - Used and Total`: the chart shows `no data` when the project does not have GKE GPU metrics. To fix the issue, update the query to provide default values 0: `outer_join 0, 0` so that when either left (GCE GPU metrics) or right (GKE) is missing, using 0s to do the join; 

For `Memory Usage % by Model`: fix the typo in the query: for this query, we want to union GCE and GKE used memory first, and union **GCE** and GKE total memory, and finally divide to get the percentage.


LujieDuan:
cc @suffiank

EvanSimpson (approved):

PR #654, opened 2023-10-26
Adds GKE Alerts for nginx-ingress, gke-ingress, http, and gmp

huygaa11:


alerts/google-gke/http-gmp-server-errors-entire-workload.v1.json
```diff
@@ -0,0 +1,33 @@
+{
+  "displayName": "Server Error (5xx) is high. (GKE Workload: ${WORKLOAD_NAME})",
```
cocosheng, line 2: Is it intended to have same displayName for two templates?

huygaa11: Yes. We have server error and latency alerts for workloads per each different type of GKE loadbalancers and ingresses.

cocosheng (approved):

PR #674, opened 2024-01-16
RabbitMQ Prometheus dashboard layout updates

algchoo:
### Description
This pull request updates the RabbitMQ Prometheus integration dashboard.

### Changes
* [updated layout for rabbitmq prometheus dashboard](https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/pull/674/commits/bc3093638e6ddc87a8b8d2615e0603892e972cca)

### Updated screenshot(s)
rabbitmq-prometheus 01
rabbitmq-prometheus 02

### Additional KPIs
* n/a


dashboards/rabbitmq/rabbitmq-prometheus.json
```diff
@@ -17,199 +17,198 @@
...
       {
-        "height": 2,
+        "xPos": 12,
+        "width": 12,
+        "height": 8,
         "widget": {
+          "title": "Incoming Messages",
           "scorecard": {
             "sparkChartView": {
               "sparkChartType": "SPARK_LINE"
```
yqlu, line 46: Just double-checking, is it intentional that these have sparklines but the other 8 scorecards in the top 2 rows don't? Is this due to what the metrics convey or should we make this consistent?

yqlu (commented):
Some quick questions:
- since this increases the metric coverage by the dashboard, do we need to update the list of metrics ingested by the integration in https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/blob/master/integrations/rabbitmq/prometheus_metadata.yaml ? Or is that still a superset of what is being used here?

dashboards/rabbitmq/rabbitmq-prometheus.json
algchoo: Great question, this is the metric:
```
sum(rate(rabbitmq_channel_messages_published_total{${Cluster},${Location},${Namespace}}[60s]) * on(instance) group_left(rabbitmq_cluster) rabbitmq_identity_info{${Cluster},${Location},${Namespace}})
```

I believe it is supposed to represent the rate at which messages are published to **channels** in the past 60 seconds, but this query is confusing me a bit by taking the sum of a rate and multiplying it by `on(instance)`, I'm not sure how to make sense of that.

Regardless of all that, we can drop the spark lines in favor of consistency, unless we feel that spikes in messages would be represented in a favorable way by the spark lines? I think that for someone looking at this card, if they see a large number, but the spark line is semi-flat, it'd be an indicator that things are at some kind of normal, whereas if we see spikes, that may indicate things are deviating from normal in their environment. If I'm wrong about the query, then perhaps these spark lines would only create confusion?

yqlu: OK, I see the value of the sparklines, but your justification would also apply to the other scorecards like "Ready Messages" and "Acknowledged Messages", correct? Basically, let's choose to add sparklines to all 4 scorecards about messages, or none of them -- and then update the screenshot.

yqlu (approved):

yqlu (approved):

PR #699, opened 2024-02-13
adding uptime-checks.v1.json

nikitajainn-google:


huygaa11 (changes requested):

alerts/google-gke/uptime-checks.v1.json
```diff
@@ -0,0 +1,30 @@
+{
+  "displayName": "${CLUSTER_NAME}/${WORKLOAD_NAME} GKE Load Balancer Check uptime failure"
+  "documentation": {},
+  "userLabels": {},
```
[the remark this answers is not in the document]
nikitajainn-google: added

alerts/google-gke/uptime-checks-workload.v1.json
```diff
@@ -0,0 +1,37 @@
+{
+  "displayName": "${CLUSTER_NAME}/${WORKLOAD_NAME} GKE Load Balancer Check uptime failure"
+  "documentation": {},
+  "userLabels": {
+    "workload_name": "${WORKLOAD_NAME}",
```
jdurelli, line 5: If we want to match all the environment variables set from the Workload Details Observability tab

workload_type = gke_deployment
location = ${LOCATION}
project_id = ${PROJECT_ID}
namespace = ${NAMESPACE}

alerts/google-gke/uptime-checks-workload.v1.json
```diff
@@ -0,0 +1,37 @@
+{
+  "displayName": "${CLUSTER_NAME}/${WORKLOAD_NAME} GKE Load Balancer Check uptime failure"
```
jdurelli, line 2: I think this is fine for now but note we can technically support uptime checks for ingress via URL, so we may want to consider passing in the "Load Balancer" part of the display name

alerts/google-gke/uptime-checks-workload.v1.json
```diff
@@ -0,0 +1,37 @@
...
+  "conditions": [
+    {
+      "conditionThreshold": {
+        "aggregations": [
+          {
+            "alignmentPeriod": "1200s",
+            "perSeriesAligner": "ALIGN_NEXT_OLDER",
+            "crossSeriesReducer": "REDUCE_COUNT_FALSE",
+	    "groupByFields": [
+              "resource.label.*"
```
jdurelli, line 18: Our current alert policy lists these out, but if we want to support ingress + load balancer through the same policy template then I think this is fine

alerts/google-gke/uptime-checks-workload.v1.json
```diff
@@ -0,0 +1,37 @@
...
+            ],
+          }
+        ],
+        "comparison": "COMPARISON_GT",
+        "duration": "${UPTIME_DURATION}",
+        "filter": "metric.type=\"monitoring.googleapis.com/uptime_check/check_passed\" AND metric.label.check_id=\"${UPTIME_CHECK_ID}\" AND resource.type=\"k8s_service\"",
```
jdurelli, line 24: Same thing for ingress... passing in the resource.type instead of hardcoding

alerts/google-gke/uptime-checks-workload.v1.json
```diff
@@ -0,0 +1,37 @@
...
+          "count": 1
+        }
+      }
+    }
+  ],
+  "alertStrategy": {
+    "autoClose": "604800s"
+  },
+  "combiner": "OR",
+  "enabled": true,
```
jdurelli, line 36: Do we support notification channels or severity? Or are they getting passed in some other way?

alerts/google-gke/uptime-checks-workload.v1.json
```diff
@@ -0,0 +1,37 @@
+{
+  "displayName": "${CLUSTER_NAME}/${WORKLOAD_NAME} GKE Load Balancer Check uptime failure"
+  "documentation": {},
+  "userLabels": {
+    "workload_name": "${WORKLOAD_NAME}",
+    "cluster": "${CLUSTER_NAME}",
+    "uptime_check_id": "${UPTIME_CHECK_ID}",
+  },
+  "conditions": [
```
jdurelli, line 9: Missing conditions.displayName -> "Failure of ${alertPolicy.displayName}"

jdurelli (approved):

alerts/google-gke/uptime-checks-workload.v1.json
nikitajainn-google: currently unsure how notification channels are getting passed into other GKE recommended alerts, but since the dropdown is there in the recommended alerts UI, I'm assuming they must be getting passed in. I've added the field

alerts/google-gke/uptime-checks-workload.v1.json
nikitajainn-google: ack, passing in the resource type

alerts/google-gke/uptime-checks-workload.v1.json
nikitajainn-google: not sure how we'll populate the alert policy display name, i think it would make sense to put the uptime check name here?

huygaa11 (approved):

alerts/google-gke/uptime-checks-workload.v1.json
nikitajainn-google: not sure if i'm missing something, where are these user labels getting populated in the workload details tab?

PR #700, opened 2024-02-14
Apache Prometheus dashboard KPIs and layout updates

mrsillydog:
## Description

This PR updates the layout and includes new KPIs for the Apache Prometheus dashboard.

## Changes

- Updated KPIs and layout for the Apache Prometheus dashboard
- Updated the Apache Prometheus dashboard screenshot
- Updated the metadata and README description for the Apache Prometheus dashboard to include the new metrics

## Updated screenshot
image

## Additional KPIs

- Apache Version
- Server Status
- Uptime (seconds)


dashboards/apache/apache-prometheus-overview.json
```diff
@@ -17,227 +18,390 @@
...
+        "width": 16,
+        "height": 8,
+        "widget": {
+          "title": "Apache Version",
+          "scorecard": {
+            "blankView": {},
+            "thresholds": [],
+            "timeSeriesQuery": {
+              "outputFullDuration": true,
+              "prometheusQuery": "apache_version{${Cluster},${Location},${Namespace}}",
```
pintofloyed, line 34: In case there are multiple apache instances in a single project, this would not be helpful to show versions of all instances. Same for server status and uptime

algchoo: That makes sense, what if we do the following:
1. Changes the version to connections and have it be a sum of all connections
```
sum(apache_connections{state="total",noop_filter=~".*",noop_filter=~".*",noop_filter=~".*"})
```
2. Change the server status to be a calculation that represents whether or not 1 or more servers are down
```
sum(apache_up{noop_filter=~".*",noop_filter=~".*",noop_filter=~".*"}) / count(apache_up{noop_filter=~".*",noop_filter=~".*",noop_filter=~".*"})
```
3. Change the uptime widget to a top list

Would these changes make the dashboard more useful with several apache instances?

I can't decide if it makes more sense for the status widget to be a regular table, so that you could see which instances are down, or if it makes more sense to just bring awareness to the fact that one or more are down.

Screenshot 2024-04-16 at 4 31 14 PM

pintofloyed: 1. Total connections looks good
2. I think two scorecards one for servers up and another for servers down should give enough info about the server status.
3. An uptime list works here. The list should map the apache instance name to it's uptime.

algchoo: @pintofloyed Thanks for the feedback!

I've set the default labels for uptime to be `pod` and the latest value. There is an `instance` label, but it's just `<POD_NAME>:prometheus` so I thought `pod` made more sense here.

The query for servers that are down is returning no value, which I think is fine if we consider that I only have one instance and it is up lol. It would be nicer to see a `0` there instead of `-`, but I'm not sure if that's feasible.

Screenshot
Screenshot 2024-04-18 at 11 40 11 AM

yqlu: (drive by comment)

Does "<query> or 0" in PromQL help display a 0 in the scorecard instead of a "-"?

algchoo: @yqlu @pintofloyed I pushed up the second option that gives a 0 on the Servers Down chart, for testing purposes. If you'd like it changed to something else, let me know. Hope you're having a great day!

pintofloyed: Can you try this syntax "or vector(0)"?

algchoo: @pintofloyed Works like a charm, thanks!

pintofloyed (approved):

PR #723, opened 2024-02-27
Memcached Prometheus dashboard KPI and layout updates

algchoo:
### Description
This PR updates the layout and includes new KPIs for the Memcached Prometheus dashboard.

### Changes
* [updated readme, metadata, and prometheus metadata to reflect additional metric](https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/commit/f4314849b006d77d1ca3d98fe810c06df92dbe0d)
* [updated dashboard with new layout and kpi metric](https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/commit/bffb0373d8bb069a94226d64a8cee20f6142fc0e)
* [updated screenshot](https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/pull/723/commits/e62b303ef605346e9ac48cc05f572cd7c8226a56)

### Updated screenshot
memcached-prometheus

### Additional KPIs
* Up/Down


dashboards/memcached/memcached-prometheus.json
```diff
@@ -147,32 +173,42 @@
...
               },
               {
+                "breakdowns": [],
+                "dimensions": [],
+                "measures": [],
                 "plotType": "LINE",
                 "targetAxis": "Y1",
                 "timeSeriesQuery": {
-                  "prometheusQuery": "irate(memcached_written_bytes_total{${Cluster},${Location},${Namespace}}[10m])"
+                  "prometheusQuery": "irate(memcached_written_bytes_total{${Cluster},${Location},${Namespace}}[10m])",
```
yqlu, line 193: Looks like this already existed before, but can we make the read vs written bytes queries consistent? The read one (several lines above this) has a `sum()` but the written one doesn't -- so if there are multiple memcached instances in the same cluster, the former will aggregate all of them to produce a single time series, but the latter would produce one time series per memcached instance.

Either seems fine, but we should make it consistent.

yqlu (changes requested):
Looks great, thanks! Just one comment.

dashboards/memcached/memcached-prometheus.json
```diff
@@ -1,5 +1,5 @@
 {
-  "category": "CUSTOM",
+  "displayName": "Memcached Prometheus Overview - KPI/LAYOUT UPDATES",
```
yqlu, line 2: Missed this earlier -- drop " - KPI/LAYOUT UPDATES" please

yqlu (commented):
Just one more change and then this LGTM!

dashboards/memcached/memcached-prometheus.json
algchoo: Sorry about that, should be good to go.

yqlu (approved):

PR #728, opened 2024-02-29
Added reservation utilization alerts to google-gce.

ryannk:
Adding template alerts to monitor reservation utilization.

- When a reservation is at high usage customers might risk running out of capacity.
- When a reservation is at low usage customers will be billed even though they have no workload running.


cocosheng (approved):

cocosheng:
please fix test. thank you

PR #748, opened 2024-03-11
Golden signal update

aburhan:


aburhan:
@yqlu can you ptal?

dashboards/google-kubernetes-engine/gke-optimization-signals.json
```diff
@@ -222,28 +223,126 @@
...
               "unitOverride": ""
             }
-          },
-          "title": "Spot Discount Node"
+          }
+        }
+      },
+      {
+        "widget": {
+          "title": "Demand-Based Downscaling",
```
yqlu, line 247: Both new charts (Demand-Based Downscaling and Spot Node Usage) have issues:
- they don't filter by the template variables
- when filters are applied such that there is only one cluster being shown, the chart value doesn't match the scorecard value
- This chart is summing allocatable_utilization while the scorecard above is averaging allocatable_utilization - which one do you want?

yqlu (changes requested):

dashboards/google-kubernetes-engine/gke-optimization-signals.json
```diff
@@ -145,32 +146,32 @@
...
       },
       {
-        "height": 5,
+        "xPos": 24,
+        "yPos": 8,
+        "width": 12,
+        "height": 20,
         "widget": {
-          "title": "AVG MEM Request vs AVG MEM Allocatable",
+          "title": "Cluster Bin Packing Memory",
```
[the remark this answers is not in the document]
aburhan: update

dashboards/google-kubernetes-engine/gke-optimization-signals.json
aburhan: updated

dashboards/google-kubernetes-engine/gke-optimization-signals.json
```diff
@@ -222,28 +223,126 @@
...
+        "width": 24,
+        "xPos": 24,
+        "yPos": 28
       }
     ]
-  }
+  },
+  "labels": {},
+  "etag": "c5d014b00f624cd6a0097ea6a4c7a11a",
+  "name": "projects/1034414536999/dashboards/41dc62cf-2581-4b16-b17f-c63e44c6ceba",
```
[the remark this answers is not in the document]
aburhan: removed

dashboards/google-kubernetes-engine/gke-optimization-signals.json
```diff
@@ -210,7 +209,7 @@
                 "plotType": "LINE",
                 "targetAxis": "Y1",
                 "timeSeriesQuery": {
-                  "timeSeriesQueryLanguage": "{ fetch k8s_container::kubernetes.io/container/cpu/request_cores;\n fetch k8s_node::kubernetes.io/node/cpu/allocatable_cores\n} \n| ${project_id}\n| ${location}\n| ${cluster_name}\n| join\n| group_by [project_id, location, cluster_name], sum(val(0)) / sum(val(1))",
+                  "timeSeriesQueryLanguage": "{ fetch k8s_container::kubernetes.io/container/cpu/request_cores\n| filter\n    resource.namespace_name !~ '(kube|istio|gatekeeper|gke|gmp|gke-gmp)-system|gke-managed-filestorecsi|gke-mcs'\n| ${project_id}\n| ${location}\n| ${cluster_name}\n    | group_by [], .aggregate()\n    ;\n fetch k8s_node::kubernetes.io/node/cpu/allocatable_cores\n| ${project_id}\n| ${location}\n| ${cluster_name}\n | group_by [], .aggregate()\n} \n| join\n| value val(0) / val(1)\n| scale '%'",
```
yqlu, line 212: If you make both the numerator and denominator `| group_by [resource.cluster_name, resource.location, resource.project_id]` as follows

Screenshot 2024-03-18 at 10 21 15 AM

You get the best of both worlds:
- when there are multiple clusters, you get one time series per cluster
- each time series is automatically labelled by the cluster name
- when the filters are drilled down to a single cluster, the single time series value matches the scorecard

And you should do this for Cluster Bin Packing Memory as well

dashboards/google-kubernetes-engine/gke-optimization-signals.json
yqlu: It looks like you've updated it to use `aggregate()` which is actually a sum, which seems wrong?

If I take away the aggregate() I see the individual nodes have allocatable utilization as follows:
Screenshot 2024-03-18 at 10 25 55 AM

But it's wrong to sum this percentage over all of the nodes and say that "allocatable utilization" for the cluster is 927%...
Screenshot 2024-03-18 at 10 27 53 AM

While mean feels more correct:
Screenshot 2024-03-18 at 10 28 47 AM

dashboards/google-kubernetes-engine/gke-optimization-signals.json
```diff
@@ -17,83 +17,83 @@
...
+        "height": 8,
         "widget": {
+          "title": "Demand-Based Downscaling",
           "scorecard": {
             "blankView": {},
             "thresholds": [],
             "timeSeriesQuery": {
-              "timeSeriesQueryLanguage": "fetch k8s_node\n| ${project_id}\n| ${location}\n| ${cluster_name}\n| metric 'kubernetes.io/node/cpu/allocatable_utilization'\n| group_by [],.mean\n| scale '%'",
+              "outputFullDuration": true,
+              "timeSeriesQueryLanguage": "fetch k8s_node\n| ${project_id}\n| ${location}\n| ${cluster_name}\n| metric 'kubernetes.io/node/cpu/allocatable_utilization'\n| group_by 1m,\n    [value_allocatable_utilization_mean: mean(value.allocatable_utilization)]\n| every 1m\n| group_by [resource.project_id, resource.location, resource.cluster_name],\n    [value_allocatable_utilization_mean_aggregate:\n       aggregate(value_allocatable_utilization_mean)]\n| value\n    [value_allocatable_utilization_mean_aggregate_scale:\n       scale(value_allocatable_utilization_mean_aggregate, '%')]",
```
yqlu, line 50: Same comment as the Demand-Based Downscaling XY chart below (the aggregate() should be a mean())

But in addition, this should be `group_by []` so that we reduce it to a single line. When I test this out in a project with multiple GKE clusters I get the following error:
Screenshot 2024-03-18 at 10 30 43 AM

yqlu (changes requested):

PR #750, opened 2024-03-12
fix: import.sh replace grep -P with GNU param expansion for MacOS support

bwplotka:
This makes import.sh usable on default MacOS terminal.

Without this change import.sh was erroring out:

```
To upload these dashboard(s) manually, you can run:
./upload.sh reports/2024-3-12/08:07:43/ <PROJECT_ID> END

grep: invalid option -- P
usage: grep [-abcdDEFGHhIiJLlMmnOopqRSsUVvwXxZz] [-A num] [-B num] [-C[num]]
	[-e pattern] [-f file] [--binary-files=value] [--color=when]
	[--context[=num]] [--directories=action] [--label] [--line-buffered]
	[--null] [pattern] [file ...]
```

This is due to MacOS shipping [BSD version of grep without Perl option.](https://stackoverflow.com/questions/77662026/grep-invalid-option-p-error-when-doing-regex-in-bash-script#comment136915500_77662026)

Replacing to more std way of doing this should support more envs.

### DIFF
--- a/scripts/dashboard-importer/import.sh
+++ b/scripts/dashboard-importer/import.sh
@@ -35,7 +35,8 @@ then
   echo "Conversion complete. Proceeding to upload..."
 
   # Extract output directory from console output
-  DIRECTORY=$(grep -Po '(reports.*/)' <<< "$CONVERSION_OUTPUT" | sed -n '2p')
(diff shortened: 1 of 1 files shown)


stevezease (approved):
This is very clean! Thank you for making this fix!

PR #765, opened 2024-05-17
5m alignment

johnbryan:
Adjust scorecard intervals from 1m to 5m. 1m was prone to volatility/partial data, and 5m should give a more accurate represenation of the current value.


yqlu (commented):
Can you tackle the "Avg CPU Usage / Request" scorecard (line 361 in gke-compute-resources-workload-view.json) as well?

johnbryan:
done

yqlu (approved):

PR #770, opened 2024-06-11
Updates GKE DCGM dashboard with latest revision

huygaa11:
The dashboard uses same set of metrics. So, backward compatible


johnbryan:
We're ok removing the `kubernetes_io:container_accelerator_...` metrics?

johnbryan (approved):

PR #817, opened 2024-08-21
Update README.md

WadeGulbrandsen:
Added troubleshooting step to verify the metric is in the Prometheus domain

### DIFF
--- a/scripts/dashboard-importer/README.md
+++ b/scripts/dashboard-importer/README.md
@@ -120,6 +120,12 @@ We recommend the following steps for debugging tiles with no data:
     inactive. You can do so by searching for the query's metric in the
     [Metrics Explorer](https://console.cloud.google.com/monitoring/metrics-explorer).
     If the metric doesn't exist, you will need to find an alternative metric.
+1.  Confirm that the metric is in the Prometheus domain. Only metrics in the Prometheus
+    domain can be converted. To verify the domain you can search for the metric in the
+    [Metrics Explorer](https://console.cloud.google.com/monitoring/metrics-explorer) and check
+    if the metric starts with `prometheus.googleapis.com`. If the metric is not in the
+    Prometheus domain you will need to find an alternative metric that is
+    [collected for Prometheus](https://cloud.google.com/stackdriver/docs/managed-prometheus#gmp-data-collection).
 1.  Confirm that the label matchers are valid.
     To validate this, first look at the query and determine if there are any
     hard equality matchers. For example, in the query `kube_pod_status_ready{container='foo'}`,
--- a/scripts/dashboard-importer/README.md
+++ b/scripts/dashboard-importer/README.md
@@ -120,8 +120,9 @@ We recommend the following steps for debugging tiles with no data:
     inactive. You can do so by searching for the query's metric in the
     [Metrics Explorer](https://console.cloud.google.com/monitoring/metrics-explorer).
     If the metric doesn't exist, you will need to find an alternative metric.
-1.  Confirm that the metric is in the Prometheus domain. Only metrics in the Prometheus
-    domain can be converted. To verify the domain you can search for the metric in the
+1.  Confirm that the metric is in the Prometheus domain. Cloud Monitoring by default looks for
+    Prometheus metrics in a PromQL query under the Prometheus domain. To verify the domain you
+    can search for the metric in the
     [Metrics Explorer](https://console.cloud.google.com/monitoring/metrics-explorer) and check
     if the metric starts with `prometheus.googleapis.com`. If the metric is not in the
     Prometheus domain you will need to find an alternative metric that is


scripts/dashboard-importer/README.md
```diff
@@ -120,6 +120,12 @@ We recommend the following steps for debugging tiles with no data:
     inactive. You can do so by searching for the query's metric in the
     [Metrics Explorer](https://console.cloud.google.com/monitoring/metrics-explorer).
     If the metric doesn't exist, you will need to find an alternative metric.
+1.  Confirm that the metric is in the Prometheus domain. Only metrics in the Prometheus
+    domain can be converted. To verify the domain you can search for the metric in the
+    [Metrics Explorer](https://console.cloud.google.com/monitoring/metrics-explorer) and check
+    if the metric starts with `prometheus.googleapis.com`. If the metric is not in the
+    Prometheus domain you will need to find an alternative metric that is
+    [collected for Prometheus](https://cloud.google.com/stackdriver/docs/managed-prometheus#gmp-data-collection).
```
stevezease, line 128: Thank you for creating this change request!
I've been meaning to make a related documentation update. 
The importer doesn't actually convert metrics. Metrics coming in empty is typically because the metrics itself has no data coming in (is inactive, covered by the previous debug step) or because the dashboard itself does not use a prometheus data source (metrics are not from prometheus).

We had added a new line addressing the second point in our [official docs](https://cloud.google.com/monitoring/dashboards/import-grafana-dashboards):`You can only import dashboards that use PromQL expressions and Prometheus data sources. Dashboards that use other forms of querying or data sources might not import successfully.`

Would it cover the case you experienced, if I added a similar debug step as the first item?
something like:
`Ensure that your dashboard uses PromQL expressions and Prometheus data sources. Dashboards that use other forms of querying or data sources might not import successfully.`

WadeGulbrandsen: I think we need to define `Prometheus data sources` better. The issue for my customer was that they had their metrics in their own Prometheus and was exporting that to GCP but the exporter they used sent them as custom metrics in the `workload.googleapis.com` domain. So from their perspective the data source was Prometheus but not what we mean by a Prometheus data source (resource type `Prometheus Target` in the `prometheus.googleapis.com` domain)

stevezease: Ah okay, I see, this appears to be a slightly different troubleshooting case. It seems the customer likely had promql based queries which look for metrics under prometheus.googleapis.com, whereas their metrics has been exported under the workload.googleapis.com domain

Happy to get this change in. In this case, I think the main tweak I would make here is regarding the second sentence, since the converter doesn't modify the metrics.

Maybe something along the lines of:
```suggestion
1.  Confirm that the metric is in the Prometheus domain. Cloud Monitoring by default looks for Prometheus metrics in a PromQL query under the Prometheus domain. To verify the domain you can search for the metric in the
    [Metrics Explorer](https://console.cloud.google.com/monitoring/metrics-explorer) and check
    if the metric starts with `prometheus.googleapis.com`. If the metric is not in the
    Prometheus domain you will need to find an alternative metric that is
    [collected for Prometheus](https://cloud.google.com/stackdriver/docs/managed-prometheus#gmp-data-collection).

WadeGulbrandsen: Sounds good to me. Made the change in a new commit.

lyanco: Hey Wade - I think this issue is different than what's described here.

In this scenario, your customer is sending Prometheus metrics to the wrong domain in Cloud Monitoring. Sending metrics to `workload.googleapis.com` is 20x more expensive than GMP, and results in a really poor user experience (as you can see, PromQL doesn't work OOTB).

So really what we need to add here is a note saying that when you move metrics from local Prometheus to Cloud Monitoring, you must use one of the approved collection options for Google Cloud Managed Service for Prometheus. In which case everything will work automatically and dashboard conversion will reference things correctly.

stevezease: Hey Wade, spoke with Lee, and we ended wanted something that more directly points customers to root of the issue in that metrics are not properly being ingested. 

We came up with the following updated troubleshooting step text to try to capture all the points we had talked about:

`
Ensure your Prometheus metrics are ingested using one of the approved collection options for [Google Cloud Managed Service for Prometheus](https://cloud.google.com/stackdriver/docs/managed-prometheus). When using the Query Builder in Cloud Monitoring's [Metrics Explorer](https://console.cloud.google.com/monitoring/metrics-explorer), you can find properly ingested Prometheus metrics under the [prometheus.googleapis.com](http://prometheus.googleapis.com/) domain. Metrics in this domain will be correctly referenced by converted dashboards.
`

Would updating to this text work for you? 
Thank you!

WadeGulbrandsen: That looks great!

stevezease: Updated and merged here! 
https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/pull/826

stevezease:
We went with the comment that we had discussed:
https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/pull/826

PR #821, opened 2024-09-06
Fix golang dashboard images

EvanSimpson:
The diff view is still broken here, but note that the file size is restored, and use the overflow menu "View File" option to verify that the images are restored.


yqlu (approved):

bwplotka:
Thanks!

PR #839, opened 2024-10-10
Add Immersive Stream dashboard

SamiSousa:


dashboards/immersive-stream/metadata.yaml
```diff
@@ -0,0 +1,7 @@
+sample_dashboards:
+  -
+    category: Immersive Stream
+    id: StreamInstance
```
EvanSimpson, line 4:
```suggestion
    id: immersive-stream
```

This needs to match the file names

EvanSimpson (changes requested):

dashboards/immersive-stream/metadata.yaml
SamiSousa: I picked this ID to match the dashboard ID for the system dashboards. Updated to match the file names instead; I don't think there will be any issues with having the IDs be different.

EvanSimpson (approved):

PR #867, opened 2024-10-24
Update BigQuery Alert Templates

atsao72:
Update the background-metadata-cache-slot-usage alert to be a PromQL query that uses a dynamic threshold based on the number of assigned slots to that reservation. This means this alert will no longer take a static threshold. 

Also, add a new alert that targets a single reservation. This will be used on the capacity management page to directly generate an alert from a chosen reservation.


atsao72:
image

cocosheng (approved):

PR #869, opened 2024-10-31
Add column metadata index staleness alert

atsao72:
This alerts if the column metadata index staleness exceeds twice the existing average of the staleness.

It does so by taking the average over 4hour sliding windows, and comparing with the 4 hour sliding window from 4 hours ago.


atsao72:
image

PR #905, opened 2025-01-23
argo-workflows: Update metrics to match upstream renaming

pnasrat:
See: https://argo-workflows.readthedocs.io/en/latest/upgrading/#metrics-changes
Signed-off-by: Pris Nasrat <pris.nasrat@chainguard.dev>

### DIFF
--- a/dashboards/argo-workflows/argo-workflows-prometheus.json
+++ b/dashboards/argo-workflows/argo-workflows-prometheus.json
@@ -30,7 +30,7 @@
             "blankView": {},
             "thresholds": [],
             "timeSeriesQuery": {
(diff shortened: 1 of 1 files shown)


yqlu:
HI @pnasrat , sorry for the late review.

Can you also update https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/blob/master/integrations/argo-workflows/prometheus_metadata.yaml ? I think we want to make two changes:
- update the metrics exported from argo
- update the minimum_supported_version from 3.4.3

This metadata is ingested for the integrations portal within GCP that links to the dashboard you're updating:
image

PR #911, opened 2025-02-04
Add Jetstream dashboard and hidden integration

yqlu:
Developed at go/jetstream-dashboard-dev-notes. We'll validate this internally before setting the integration status from HIDDEN to GA.


dashboards/jetstream/metadata.yaml
```diff
@@ -0,0 +1,9 @@
+sample_dashboards:
+  -
+    category: Jetstream
+    id: jetstream-prometheus
+    display_name: Jetstream Prometheus Overview
+    description: "This dashboard has charts displaying request rate, latency, queue sizes and other metrics for Jetstream inference."
+    related_integrations:
+      - id: jetstream
```
EvanSimpson, line 8: FYI dashboard and integration will deploy separately

dashboards/jetstream/jetstream-overview.json
```diff
@@ -0,0 +1,1004 @@
+{
+  "displayName": "JetStream Prometheus Overview",
```
EvanSimpson, line 2: This is the only place the s in JetStream is capitalized, should this + images be updated, or yaml files?

yqlu: Good catch! Based on:
- https://cloud.google.com/kubernetes-engine/docs/tutorials/serve-gemma-tpu-jetstream
- https://github.com/AI-Hypercomputer/JetStream

It looks like "JetStream" is right, so I've fixed everything else to match.

dashboards/jetstream/metadata.yaml
yqlu: Ack. I don't think it's a big deal because:
- the integration is being added as HIDDEN -- I'll wait for both to propagate fully, validate it internally and publish the corresponding docs before we mark it as GA
- I believe our FE code fails gracefully if either deploys before the other

EvanSimpson (approved):

PR #914, opened 2025-02-08
Add jetstream documentation yaml

yqlu:


integrations/jetstream/documentation.yaml
```diff
@@ -0,0 +1,40 @@
+exporter_type: included
+app_name_short: JetStream
+app_name: {{app_name_short}}
+app_site_name: {{app_name_short}}
+app_site_url: https://github.com/AI-Hypercomputer/JetStream
+exporter_name: {{app_name_short}}
+exporter_pkg_name: {{app_name_short}}
+exporter_repo_url: https://github.com/AI-Hypercomputer/JetStream/blob/main/docs/observability-prometheus-metrics-in-jetstream-server.md
+additional_prereq_info: |
+  {{app_name_short}} exposes Prometheus-format metrics on version > 0.2.4. If no such image is publicly hosted, follow the instructions [here](https://github.com/AI-Hypercomputer/jetstream-pytorch/tree/main/docker/jetstream-pytorch-server) to build your own.
```
yqlu, line 10: @mtabasko we'll need to tweak this wording.

Basically there is currently no such image (the latest is v0.2.4, which does not expose metrics), so you have to build your own. The Jetstream team let us know that v0.2.5 is in the works, however. I don't suppose we can say "at the time of writing, ..." in here 😉

johnbryan (approved):
LGTM. Leaving up to you if you want to address your comment with mtabasko before merging.

integrations/jetstream/documentation.yaml
```diff
@@ -0,0 +1,40 @@
+exporter_type: included
+app_name_short: JetStream
+app_name: {{app_name_short}}
+app_site_name: {{app_name_short}}
+app_site_url: https://github.com/AI-Hypercomputer/JetStream
+exporter_name: {{app_name_short}}
+exporter_pkg_name: {{app_name_short}}
```
mtabasko, line 7: exporter_pkg_name isn't used for "included" types.

integrations/jetstream/documentation.yaml
mtabasko: So, if I understand correctly, right now, you just have to build your own. It seems really weird to tell them to look for something we know doesn't exist (and if JetStream doesn't get it into 0.2.5, then our docs will be wrong).

We probably have to say something like

To collect Prometheus-format metrics from {{app_name_short}},  you must first [build and upload a {{app_name_short}} PyTorch server image](https://github.com/AI-Hypercomputer/jetstream-pytorch/tree/main/docker/jetstream-pytorch-server).

(When 0.2.5 is available, we can amend this to say "from versions of {{app_name_short}} older than v0.2.5" or something.

mtabasko (changes requested):

integrations/jetstream/documentation.yaml
```diff
@@ -0,0 +1,39 @@
+exporter_type: included
+app_name_short: JetStream
+app_name: {{app_name_short}}
+app_site_name: {{app_name_short}}
+app_site_url: https://github.com/AI-Hypercomputer/JetStream
+exporter_name: {{app_name_short}}
+exporter_repo_url: https://github.com/AI-Hypercomputer/JetStream/blob/main/docs/observability-prometheus-metrics-in-jetstream-server.md
```
mtabasko, line 7: Forgot about this: We should also add a variable for the GKE setup doc:

gke_setup_url: /kubernetes-engine/docs/tutorials/serve-gemma-tpu-jetstream

(That's my guess as to the value. Please correct if wrong!!)

PR #915, opened 2025-02-08
Implement suggestions for Triton and vLLM dashboards

yqlu:
1. from go/vllm-dashboard-dev-notes : suggestion to add a new scorecard in the top level for # waiting tasks as a signal of saturation
2. from go/triton-dashboard-dev-notes : report that GMP latency metrics have a unit of "1" so the TimeSeriesFilter charts were extracting the wrong unit -- converting this to PromQL and specifying a unit override fixes this.


dashboards/nvidia-gpu/nvidia-triton-prometheus.json
```diff
@@ -616,81 +600,43 @@
         "width": 24,
         "height": 16,
         "widget": {
-          "title": "Inference latency by percentile",
           "xyChart": {
             "dataSets": [
```
johnbryan, line 604: consider keeping the legend templates here?

johnbryan (approved):

dashboards/nvidia-gpu/nvidia-triton-prometheus.json
yqlu: Good call, done (and updated the screenshot too). I don't remember why legend templates don't show up in the UI for PromQL queries but you can set them in the JSON.

PR #936, opened 2025-02-28
add integration for kuberay

xiangshen-dk:


integrations/kuberay/prometheus_metadata.yaml
```diff
@@ -0,0 +1,96 @@
...
+        value_type: DOUBLE
+      - name: prometheus.googleapis.com/ray_node_network_receive_speed/gauge
+        prometheus_name: ray_node_network_receive_speed
+        kind: GAUGE
+        value_type: DOUBLE
+      - name: prometheus.googleapis.com/ray_node_network_send_speed/gauge
+        prometheus_name: ray_node_network_send_speed
+        kind: GAUGE
+        value_type: DOUBLE
+    install_documentation_url: https://cloud.google.com/kubernetes-engine/docs/add-on/ray-on-gke/how-to/enable-ray-on-gke
```
yqlu, line 96: It sounds like we actually have two docs:
1. https://cloud.google.com/kubernetes-engine/docs/add-on/ray-on-gke/how-to/enable-ray-on-gke for how to set up Kuberay
2. a doc in stackdriver for how to use GMP to scrape metrics for Kuberay (which is what `documentation.yaml` will produce).

This field populates this link in the UI:
image

Which one do you want to link to? And should those docs 1 and 2 link to each other for discoverability as well?

integrations/kuberay/prometheus_metadata.yaml
```diff
@@ -0,0 +1,96 @@
+platforms:
+  - type: GKE
+    launch_stage: GA
```
yqlu, line 3: If possible I would recommend launch as HIDDEN first, test that everything works (metric detection, everything displays correctly, links are configured correctly), before setting launch_stage as GA

dashboards/kuberay/metadata.yaml
```diff
@@ -5,3 +5,6 @@ sample_dashboards:
     display_name: Kuberay Overview
     description: |-
       This dashboard has 20 charts based on [Ray prometheus metrics](https://docs.ray.io/en/latest/cluster/metrics.html). It includes charts displaying: Scheduler Task State, Active Tasks by Name, Scheduler Actor State, Active Actors by Name, Scheduler CPUs (logical slots), Object Store Memory, Scheduler GPUs (logical slots), Scheduler Placement Groups, Node CPU (hardware utilization), Node GPU (hardware utilization), Node Disk, Node Disk IO Speed, Node Memory (heap + object store), Node Out of Memory Failures by Name, Node Memory by Component, Node CPU by Component, Node GPU Memory (GRAM), Node Network, Node Count, Cluster Utilization.
+    related_integrations:
+      - id: kuberay
+        platform: GKE
```
yqlu, line 10: nit: newline at the end of the file

integrations/kuberay/prometheus_metadata.yaml
yqlu: Especially if you need to coordinate with a docs writer on the docs corresponding to the documentation.yaml

xiangshen-dk: Changed to HIDDEN.

integrations/kuberay/prometheus_metadata.yaml
xiangshen-dk: I'd like to promote the usage of the kuberay operator from GCP. If the operator is used, monitoring can be enabled to collect the metrics during the installation. If that's the case, do I still need the `documentation.yaml` file?

yqlu: The documentation.yaml page is used by the tech writers to create a doc in this directory: https://cloud.google.com/stackdriver/docs/managed-prometheus/exporters/introduction

For example, the [vLLM one](https://cloud.google.com/stackdriver/docs/managed-prometheus/exporters/vllm) in this directory assumes you already have set up vLLM, but want the information on how to observe it.

It might make sense to have docs in both places, especially for discoverability. But then you should work with the tech writers so that the docs link to each other.

xiangshen-dk: yes, I understand this part. However, Ray is a little different since we do have a supported operator for it. Anyhow, I will reach out to the tech writer. Thanks!

yqlu (approved):

PR #939, opened 2025-03-03
Rename Inference Extension to Inference Optimized Gateway

JeffLuoo:
Keep the name of ID to match oss repo name and they are for internal reference only.


JeffLuoo:
@yqlu

yqlu (approved):
Do you mind retaking the first screenshot too?

JeffLuoo:
Updated.

PR #943, opened 2025-03-12
Update the Lines of Code query to use the max value for a given reque…

louisjimenez:
This updates the query used for the Lines of Code Accepted by Day graph. This update accounts for log entries with a shared request ID possibly being emitted out of order.


louisjimenez:
@yqlu @johnbryan Would either of you be able to take a look? Thank you!

louisjimenez:
Thanks @johnbryan , updated.

johnbryan (approved):

PR #946, opened 2025-03-17
Add new dashboard "GKE TPU Node Pool Status"

jakubin:


dashboards/google-kubernetes-engine/README.md
```diff
@@ -68,4 +68,11 @@
 |Filename: [gke-dpv2-observability-drilldown-view.json](gke-dpv2-observability-drilldown-view.json)|
 |This dashboard features 6 charts providing pod-to-workload traffic data that is groupable and filterable (e.g. by workload, namespace), letting you gain more detailed insight into the traffic within your GKE cluster. Understand which workloads generate the most traffic, drill down which pods drop traffic and what's the reason, and more. [Google Managed Prometheus](https://cloud.google.com/managed-prometheus) and [Dataplane V2 metrics](https://cloud.google.com/kubernetes-engine/docs/how-to/configure-dpv2-observability) must be enabled to make use of this dashboard.|
 
+&nbsp;
+
+|GKE TPU Node Pool Status|
+|:---------------------|
+|Filename: [gke-tpu-node-pool-status.json](gke-tpu-node-pool-status.json)|
+|This dashboard has 18 charts for the related [Kubernetes metrics](https://cloud.google.com/monitoring/api/metrics_kubernetes), including metrics for the multi-host TPU node pool status and availability.|
```
JeffLuoo, line 76: I would suggest not mentioning `18 charts`, instead, just say `charts` in case we update the dashboard later and forget to update it.

jakubin: Fixed. Good that you've spotted it, because it was a copy-paste issue. This dashboard doesn't have 18 charts 😬

dashboards/google-kubernetes-engine/gke-tpu-node-pool-status.json
```diff
@@ -0,0 +1,333 @@
...
+      },
+      {
+        "height": 20,
+        "widget": {
+          "text": {
+            "content": "Get comprehensive insights into the health of your GKE TPU multi-host node pools. **The dashboard only shows data for TPU multi-host node pools.**\n\n### Charts:\n\n* **Node pool availability:**\n    * Tracks real-time availability to help you identify problem node pools\n    * 100% indicates all nodes are operational; 0% indicates an issue\n* **Node pool status:**\n    * Historical status charts for each pool to help you pinpoint status changes, especially errors\n    * 100% line indicates a sustained status\n* **Node pools by status**\n    * Historical number of node pools in each status\n    * Single pool filter shows its status timeline\n\n### How to use:\n\n* **Filter:** Use cluster-level filters (cluster_name, location, project_id) to focus on your GKE resource\n* **Identify issues:** Review the \"Node pool availability\" chart for dips below 100%\n* **Investigate:** Examine \"Node pool status history\" charts for problem pools to understand status changes\n* **Analyze:** Use the node pool name filter to isolate and analyze individual node pool behavior over time",
```
johnbryan, line 291: This refers to `"node pool status history" charts` in quotes but I don't see that specific string - maybe just refer to `node pool status charts`?

dashboards/google-kubernetes-engine/gke-tpu-node-pool-status.json
```diff
@@ -0,0 +1,333 @@
...
+                "minAlignmentPeriod": "60s",
+                "plotType": "LINE",
+                "targetAxis": "Y1",
+                "timeSeriesQuery": {
+                  "timeSeriesFilter": {
+                    "aggregation": {
+                      "alignmentPeriod": "60s",
+                      "crossSeriesReducer": "REDUCE_MEAN",
+                      "groupByFields": [
+                        "resource.label.\"node_pool_name\""
```
johnbryan, line 51: For all the charts, you could consider grouping by the full set of unique identifiers for a node pool - project/location/cluster/pool name. The upside of this is if there is a node pool name reused across clusters, each one will be its own line. The only downside is it can clutter chart legends. In other dashboards we've used legend templates as a workaround for this, here's an example that shows `node_pool_name (cluster_name)`: 
group-by and legend template

johnbryan (changes requested):
Overall looks good, just a couple comments

dashboards/google-kubernetes-engine/gke-tpu-node-pool-status.json
jakubin: Thanks, corrected.

dashboards/google-kubernetes-engine/gke-tpu-node-pool-status.json
jakubin: Good to know about the legend - thanks.

I have doubts if we need grouping by cluster labels in this dashboard, though. It is meant to be used to investigate issues in a single cluster. So selecting it from filters is the first step (that's also in dashboard description). As a side note, we're going to link to this dashboard with cluster pre-selected from a different page in cloud console and that is going to be the main entry point. If the dashboard is meant to be used in the context of a single cluster, I think that adding cluster name to the legend will make it harder to read. 

Unless you insist, I'd rather have it as is (I've validated its content with PM+UX+UX Writer earlier) and if make changes later if we discover that users find something confusing.

johnbryan: Makes sense, works for me!

johnbryan (approved):

PR #955, opened 2025-03-26
Switch to use ClusterPodMonitoring with secret in inference gateway

JeffLuoo:
The secret is required because the metrics endpoint is protected.


JeffLuoo:
Ref PR in the example repo: https://github.com/GoogleCloudPlatform/prometheus-engine/pull/1552

yqlu (approved):
Note that this does not automatically propagate to the docs by default.

FYI @mtabasko to update the docs as a consequence of this.

PR #957, opened 2025-04-01
Updated Cloud Run Dashboard

gxnicholas:
At a high level, there are two main changes:

1) Switch from grid layout to mosaic layout (which supports additional layout widgets) 
2) Group by Cloud Run service across all dashboard charts

Some smaller changes include:

1) Add dashboard filters for project ID, location, and service name
2) For request count metrics, use `REDUCE_SUM` instead of `REDUCE_COUNT`
3) Create tab groups for p50/p95/p99 charts
4) Add a new container instance count chart


gxnicholas:
@yqlu PTAL when you get a chance

dashboards/google-cloud-run/cloudrun-monitoring.json
```diff
@@ -1,343 +1,765 @@
...
+                      ],
+                      "perSeriesAligner": "ALIGN_RATE"
+                    },
+                    "filter": "metric.type=\"run.googleapis.com/container/memory/allocation_time\" resource.type=\"cloud_run_revision\""
                   },
-                  "filter": "metric.type=\"run.googleapis.com/container/cpu/utilizations\" resource.type=\"cloud_run_revision\"",
-                  "secondaryAggregation": {}
-                },
-                "unitOverride": "10^2.%"
+                  "unitOverride": "GiBy.s"
```
johnbryan, line 425: do you need unit overrides for cpu/memory allocation? It looks like this is the default unit https://screenshot.googleplex.com/BRkAvQzqqxU5LA2

johnbryan (changes requested):
Did you want to add anything about status code breakdowns? Maybe aggregated request rate by status, and if someone wants to see per service they use the filter at the top?

johnbryan:
Generally looks great and is a huge improvement!

dashboards/google-cloud-run/cloudrun-monitoring.json
gxnicholas: Without this override for memory allocation, I believe the unit will just be GiB as opposed to the unit of GiB.s/s which the current dashboard has

Without override:

Screenshot 2025-04-01 at 2 10 03 PM

With override: 
Screenshot 2025-04-01 at 2 11 01 PM

The same goes for CPU. It depends on whether we want the s/s.

gxnicholas:
> Did you want to add anything about status code breakdowns? Maybe aggregated request rate by status, and if someone wants to see per service they use the filter at the top?

An alternative would be to still group by service but have separate tabs for each response code class. Do you think that approach would make sense? I do think it might be odd to have one chart that is not grouped by service while the rest are. Another alternative would be to do this as a table with request count and a column containing % for each response code class.

gxnicholas:
> > Did you want to add anything about status code breakdowns? Maybe aggregated request rate by status, and if someone wants to see per service they use the filter at the top?
> 
> An alternative would be to still group by service but have separate tabs for each response code class. Do you think that approach would make sense? I do think it might be odd to have one chart that is not grouped by service while the rest are. Another alternative would be to do this as a table with request count and a column containing % for each response code class.

For now I have gone with the table approach. New screenshots including the table have been added as well.

PR #958, opened 2025-04-07
Rename inference optimized gateway to GKE Inference Gateway

JeffLuoo:
The link to the exporter example page will be updated in a follow-up.


yqlu (approved):
LGTM. Just double-checking that we are intentionally not changing the id (`gateway-api-inference-extension`) so as to not change the URLs of these integration dashboards.

PR #967, opened 2025-04-15
Update inference gateway dashboard

JeffLuoo:
- Rename some charts in the dashboard
- Include more percentiles in the histrogram metrics
- Add new metrics from inference server deployment


dashboards/gateway-api-inference-extension/inference-extension-prometheus.json
```diff
@@ -274,153 +341,197 @@
...
               {
+                "breakdowns": [],
+                "dimensions": [],
+                "legendTemplate": "",
+                "measures": [],
                 "plotType": "LINE",
                 "targetAxis": "Y1",
                 "timeSeriesQuery": {
+                  "outputFullDuration": false,
                   "prometheusQuery": "sum by (name)(avg_over_time(inference_pool_average_kv_cache_utilization[${__interval}]))*100",
```
johnbryan, line 519: totally optional / just FYI: you can use a unit of `10^2.%` too, and then you don't need `*100` in queries.

johnbryan (approved):
Changes look good!

Nit/for future reference - slight preference to take screenshots on a smaller window (say, 1500px width). They show up as a preview image in a UI flyout, and while I doubt most users are inspecting super carefully there, it's nice if they're a little more easily viewable.

PR #969, opened 2025-04-17
Add BigQuery continuous query dashboard template and screenshots.

jonahtang:


dashboards/google-bigquery/bigquery-continuous-query.json
```diff
@@ -0,0 +1,290 @@
...
+                "minAlignmentPeriod": "60s",
+                "plotType": "LINE",
+                "targetAxis": "Y1",
+                "timeSeriesQuery": {
+                  "outputFullDuration": false,
+                  "timeSeriesFilter": {
+                    "aggregation": {
+                      "alignmentPeriod": "60s",
+                      "crossSeriesReducer": "REDUCE_SUM",
+                      "groupByFields": [],
```
yqlu, line 162: Not a domain expert, but do you want to group by stage_id here given that you're doing so for other charts?

https://screenshot.googleplex.com/76vfk7s48qLzdQB

dashboards/google-bigquery/bigquery-continuous-query.json
```diff
@@ -0,0 +1,290 @@
...
+                "breakdowns": [],
+                "dimensions": [],
+                "legendTemplate": "",
+                "measures": [],
+                "plotType": "LINE",
+                "targetAxis": "Y1",
+                "timeSeriesQuery": {
+                  "outputFullDuration": false,
+                  "prometheusQuery": "delta(bigquery_googleapis_com:job_continuous_estimated_bytes_processed{monitored_resource=\"bigquery.googleapis.com/ContinuousJob\"}[5m])",
+                  "unitOverride": ""
```
yqlu, line 276: Optional: specify unit override here as well to make the y-axis nicer: https://screenshot.googleplex.com/3uYXnupCRdhbNtQ

(The backlog bytes chart already has the right units because it's in query builder, but since this is a PromQL query it drops all unit info and you have to specify it back again)

dashboards/google-bigquery/bigquery-continuous-query.json
```diff
@@ -0,0 +1,290 @@
+{
+  "displayName": "BigQuery Continuous Query Monitoring",
```
yqlu, line 2: Optional suggestions:
1. is "Continuous Query" and "Continuous Job" the same thing? I see that the metrics and resources all say "Continuous Job" but we're saying "Continuous Query" in the title
2. Consider if it's necessary to add a text widget at the top with instructions on how to use this dashboard, or a link to docs about the product.

dashboards/google-bigquery/metadata.yaml
```diff
@@ -8,3 +8,12 @@ sample_dashboards:
     related_integrations:
       - id: bigquery
         platform: GCP
+  -
+    category: BigQuery
+    id: bigquery-continuous-query
+    display_name: BigQuery Continuous Query Monitoring
+    description: |-
+      This dashboard has charts displaying BigQuery metrics for a continuous query.
```
yqlu, line 16: You want to consider if this is a dashboard for viewing metrics about a single continuous query or an overview dashboard across multiple queries (or both). It looks like the test project you're building this in has just one time series for one job_id, so there isn't a meaningful distinction.

But if it's possible to have multiple job_ids in the same project, you probably want:
1) some charts which are grouped by job_id (if it's meaningful to compare one job_id against another)
2) if some of these metrics are not super meaningful when aggregated across multiple job_ids, then you should have instructions telling the user to filter down to a single job_id first before looking at the charts.

image

See [this](https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/blob/master/dashboards/google-vertex-ai/vertex-ai-model-garden.02.png) for an example where we did this recently.

yqlu (approved):
Some optional comments

dashboards/google-bigquery/bigquery-continuous-query.json
```diff
@@ -0,0 +1,290 @@
...
+                "breakdowns": [],
+                "dimensions": [],
+                "legendTemplate": "",
+                "measures": [],
+                "plotType": "LINE",
+                "targetAxis": "Y1",
+                "timeSeriesQuery": {
+                  "outputFullDuration": false,
+                  "prometheusQuery": "min without (stage_id) (time() * 1000000 - bigquery_googleapis_com:job_continuous_output_watermark) / 1000000",
+                  "unitOverride": ""
```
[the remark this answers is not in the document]
jonahtang: Thank you, added!

dashboards/google-bigquery/bigquery-continuous-query.json
jonahtang: 1. The monitored resource is called a ContinuousJob as technically a "query" is a type of job in BigQuery. The product/feature name is also "Continuous Queries" so we have decided to stick with that in most titles and leave job for only the monitored resource name for now.
2. Added a text widget with links to docs and some information (e.g. its use primarily to monitor a single continuous query).

dashboards/google-bigquery/bigquery-continuous-query.json
jonahtang: Good call-out, this is intentional for now as we only produce backlog bytes for the input stages.

dashboards/google-bigquery/bigquery-continuous-query.json
jonahtang: Done.

dashboards/google-bigquery/metadata.yaml
jonahtang: I've added instructions in a text box to filter to a single job id.

dashboards/google-bigquery/bigquery-continuous-query.json
```diff
@@ -0,0 +1,313 @@
+{
+  "displayName": "BigQuery Continuous Query Monitoring",
+  "dashboardFilters": [],
```
yqlu, line 3: Any user can apply a job_id filter this way: https://screencast.googleplex.com/cast/NDgyNTc4MjE3NzIzNDk0NHwyM2YwZmQ1ZS02Mw but the "blue" chip signifies that it is not persistent, and will go away if you refresh the page

A slightly better UX (fewer clicks, more discoverable) is to declare a persistent filter this way: https://screencast.googleplex.com/cast/NTA5Nzc5MDkxMTgwNzQ4OHw4YjI2MzlkZC0xOA which will show up as a config in this JSON block.

yqlu (approved):
Thanks for working through all the comments, just one final point!

dashboards/google-bigquery/bigquery-continuous-query.json
jonahtang: Ah that's great, declared a persistent filter instead.

yqlu: The default is multi-select which manifests as STRING_ARRAY
Screenshot 2025-04-29 at 1 52 39 PM

Screenshot 2025-04-29 at 1 55 45 PM

Just double-checking you want that. If you unselect that in the UI, the JSON here is
```
    {
      "filterType": "RESOURCE_LABEL",
      "labelKey": "job_id",
      "stringValue": "",
      "valueType": "STRING"
    },
```

which manifests as 
Screenshot 2025-04-29 at 1 56 24 PM

jonahtang: Thanks for double checking! Yes, I think that's what we want for now as we do want to allow users (to a lesser extent) to see multiple jobs in the dashboard.

yqlu (approved):

PR #985, opened 2025-05-27
MQL->PromQL: Flink GCE Jobs Overview

mrsillydog:
This PR updates the Flink GCE Jobs Overview Dashboard to use PromQL instead of the deprecated MQL.

To prove equivalence, here are two screenshots of two different versions of the dashboard over the same time period. The former is the previous version of the dashboard, the latter the updated version of the dashboard.

MQL:
image

PromQL:
image


varun-c:
Some of the charts being updated to use PromQL are time series filter queries (not MQL). Is this intentional?

PR #989, opened 2025-05-29
PromQL: Solr

mrsillydog:
This PR updates the Solr GCE Overview Dashboard to use PromQL instead of the deprecated MQL.

Before:
image

After:
image

Conversion Issues:


mrsillydog:
This PR is being closed in favor of the more reviewable https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/pull/1040

PR #998, opened 2025-06-03
PromQL: Dataproc Job

mrsillydog:
This PR updates the Dataproc Job Monitoring Dashboard to use PromQL instead of the deprecated MQL.

To prove equivalence*, here are screenshots of two different versions of the dashboard over the same time period. The former is the previous version of the dashboard, the latter the updated version of the dashboard.

Before:
image

After:
image

*this is the first instance found where PromQL is not really able to mimic MQL's behavior. The panel that used to be `Slow jobs (>=120s) by project` (now `Slow jobs (>=64s) by project`) has some inconsistencies with MQL's behavior that we cannot address. Issues below.

Conversion Issues:
- You cannot create arbitrary buckets for histogram data in PromQL the way you can in MQL. Specifically, you cannot grab all jobs that are > 120s for the `Slow jobs` panel - you have to work with the predefined buckets PromQL gives you. See the bucket options in the screenshot below. I picked 64 seconds because 128 was not an option, and it seemed better than 256.
image

- You cannot mimic the MQL `sliding(1h)` behavior in the `Slow jobs` panel which results in data being sampled once every hour, on the hour. PromQL is opinionated about the sample rate. To prove relative equivalence outside of the sample rate, I've attached screenshots of the data values at two points for each dashboard, which are equivalent. The data in around these points is just different because of PromQL's every-minute sample rate - MQL only has those two data points, whereas PromQL has many more.

MQL, 11 AM: 
image

PromQL, 11 AM:
image

MQL, noon:
image

PromQL, noon:
image


varun-c (approved):

mrsillydog:
> > > Looking at the screenshots, the # of running jobs at 11:00 AM appears to be fewer than the # of slow jobs at 11:00 AM. Do you know why this is the case, and if we can make any changes to reduce the confusion? Is it because the slow jobs metric looks at the past 1 hour whereas the running jobs looks at the past minute?
> > 
> > 
> > @varun-c I'm a bit confused - at exactly 11 AM, both of the dashboards have a value of 49. Can you clarify what you mean?
> > image
> 
> I was comparing the value shown in the "Slow Jobs by project" chart to the value shown in the "Running jobs by project" chart. At 11 AM, the former shows 49, but the latter seems to show ~20 running jobs. This issue is present in the MQL version of the dashboard too, so I think it is fine to merge this PR. I was just curious if you knew the reason why the # of running jobs would show up as fewer than the # of slow jobs? Is it because the two charts have different aggregation windows?

I understand now, thanks - yes, I believe it's because of the aggregation window. The Running jobs by project has a window of every 1m, whereas the Slow jobs by project looks at the last hour.

varun-c:
Thanks for confirming!

PR #1000, opened 2025-06-04
PromQL: CouchDB

braxtonorban:
This PR updates the CouchDB GCE Overview Dashboard to use PromQL instead of the deprecated MQL.

Before:
image

After:
image

Conversion Issues:


dashboards/couchdb/couchdb-gce-overview.json
```diff
@@ -1,399 +1,538 @@
...
       },
       {
-        "height": 4,
+        "yPos": 64,
+        "xPos": 16,
+        "height": 16,
+        "width": 16,
         "widget": {
-          "title": "Open File Descriptors",
+          "title": "Memory % Top VMs",
```
varun-c, line 453:
```suggestion
          "title": "Memory % Top 5 VMs",
```

braxtonorban:
https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/pull/1044

PR #1001, opened 2025-06-04
PromQL: Aerospike

braxtonorban:
This PR updates the Aerospike GCE Overview Dashboard to use PromQL instead of the deprecated MQL.

Before:
image

After:
image

Conversion Issues:

Small discrepancy with

### DIFF
--- a/dashboards/aerospike/aerospike-gce-overview.json
+++ b/dashboards/aerospike/aerospike-gce-overview.json
@@ -1,260 +1,353 @@
 {
   "displayName": "Aerospike GCE Overview",
+  "dashboardFilters": [],
+  "labels": {},
   "mosaicLayout": {
-    "columns": 12,
+    "columns": 48,
     "tiles": [
       {
-        "height": 4,
+        "height": 16,
+        "width": 24,
         "widget": {
           "title": "Free Memory %",
+          "id": "",
           "xyChart": {
             "chartOptions": {
-              "mode": "COLOR"
+              "displayHorizontal": false,
+              "mode": "COLOR",
+              "showLegend": false
             },
             "dataSets": [
               {
+                "breakdowns": [],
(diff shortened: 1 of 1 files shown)


dashboards/aerospike/aerospike-gce-overview.json
```diff
@@ -1,261 +1,354 @@
...
+                "breakdowns": [],
+                "dimensions": [],
+                "legendTemplate": "",
+                "measures": [],
                 "plotType": "LINE",
                 "targetAxis": "Y1",
                 "timeSeriesQuery": {
-                  "timeSeriesQueryLanguage": "fetch gce_instance::workload.googleapis.com/aerospike.namespace.memory.free | group_by [metric.node_name, metric.namespace_name] | bottom 5, val(0)\n"
+                  "outputFullDuration": false,
+                  "prometheusQuery": "bottomk(5, avg by (node_name, namespace_name) (workload_googleapis_com:aerospike_namespace_memory_free{monitored_resource=\"gce_instance\"}))",
```
varun-c, line 166: In the screenshot, this chart has 6 time series though it is looking for bottom 5. The chart for the MQL version of the query appears to have only 5 time series. Do you know why there is a discrepancy?

braxtonorban: I'm not certain why this discrepancy occurs, but I believe it is because bottomk in PromQL returns the bottomk time series at each timestamp, while MQL strictly displays only 5 time series. In my testing so far I was not able to find a PromQL query that perfectly mimicked the MQL behavior, so I would appreciate your thoughts on the best way to handle this query!

braxtonorban:
https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/pull/1045

PR #1003, opened 2025-06-05
PromQL: Google Anthos

mrsillydog:
This PR updates the Google Anthos Dashboards to use PromQL instead of the deprecated MQL.

No screenshots are included here, since generating load on Anthos metrics is non-trivial. Conversions are best effort, but should largely be 1-1. Exception is MQL queries that UNION two queries with different monitored resources - for clarity, these queries have been split into two separate PromQL queries for the same dashboard panel.


mrsillydog:
https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/pull/1080

PR #1012, opened 2025-06-09
PromQL: Google Cloud SQL

mrsillydog:
This PR updates the Google Cloud SQL Dashboards to use PromQL instead of the deprecated MQL.

To prove equivalence*, here are screenshots of two different versions of the dashboards over the same time period. The former is the previous version of the dashboard, the latter the updated version of the dashboards.

General - Before (updated panels only):
image

General - After (updated panels only):
image

Replication - Before:
image

Replication - After:
image

Transactions - Before:
image
image

Transactions - After:
image
image

*one small exception to equivalence applies, see below.

1) In the Replication dashboard, the "Top 5 Replications with Highest Memory Usage" panel's y-axis scale is different between before and after. This is because there was something of a bug with the old version - basically, the MQL `join` resulted in five identical time series per replica, only differentiated by their `state` label - the PromQL `and on` clause does not result in this same fan out behavior. When these time series were aggregated on a per-replica basis, this resulted in the value being 5x higher than it should have been, unless I'm expected to believe that both of my recently-created replicas were already over 200% CPU usage.

### DIFF
--- a/dashboards/google-cloudsql/cloudsql-general.json
+++ b/dashboards/google-cloudsql/cloudsql-general.json
@@ -1,83 +1,100 @@
 {
+  "displayName": "Cloud SQL - General",
   "dashboardFilters": [
     {
       "filterType": "RESOURCE_LABEL",
-      "labelKey": "region"
+      "labelKey": "region",
+      "templateVariable": "",
+      "valueType": "STRING"
     },
     {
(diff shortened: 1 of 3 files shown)


mrsillydog:
Closing in favor of the more easily-reviewable https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/pull/1058

PR #1014, opened 2025-06-09
PromQL: Google Dataflow Job

mrsillydog:
This PR updates the Google Cloud SQL Dashboards to use PromQL instead of the deprecated MQL.

To prove equivalence, here are screenshots of two different versions of the dashboards over the same time period. The former is the previous version of the dashboard, the latter the updated version of the dashboards.

Before (updated panel only):
image

After (updated panel only):
image


mrsillydog:
Closing in favor of https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/pull/1059

PR #1016, opened 2025-06-09
PromQL: Google Filestore

mrsillydog:
This PR updates the Google Filestore Dashboard to use PromQL instead of the deprecated MQL.

Before (updated panels only):
image

After (updated panels only):
image

If desired, I can change the alignment of the Latest Value column to be the same as before, but I was of the opinion that it actually looked better right-aligned.


mrsillydog:
Closing in favor of https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/pull/1060

PR #1017, opened 2025-06-10
PromQL: MongoDB

braxtonorban:
This PR updates the MongoDB GCE Overview Dashboard to use PromQL instead of the deprecated MQL.

Before:
image

After: 
image

Conversion Issues:


dashboards/mongodb/mongodb-gce-overview.json
```diff
@@ -665,181 +860,231 @@
...
+                "dimensions": [],
+                "legendTemplate": "",
+                "measures": [],
                 "plotType": "LINE",
                 "targetAxis": "Y1",
                 "timeSeriesQuery": {
-                  "timeSeriesQueryLanguage": "def top_5_cpu_filtered_by_metric filter_metric =\n  fetch gce_instance\n  | { t_cpu: metric 'compute.googleapis.com/instance/cpu/utilization'; \n  t_filter_metric: metric $filter_metric }\n  | join\n  | value t_cpu.value.utilization\n  | group_by [resource.project_id, resource.zone, metric.instance_name], 1m,\n      [value_utilization_mean: mean(t_cpu.value.utilization)]\n  | top 5\n  | every 1m;\n\n@top_5_cpu_filtered_by_metric 'workload.googleapis.com/mongodb.memory.usage'\n"
+                  "outputFullDuration": false,
+                  "prometheusQuery": "topk(5, 100 *\n  avg by (project_id, zone, instance_name) (\n    avg_over_time(\n      compute_googleapis_com:instance_cpu_utilization{monitored_resource=\"gce_instance\"}[1m]\n    )\n    and\n    on(instance_id, project_id, zone)\n    (workload_googleapis_com:mongodb_memory_usage{monitored_resource=\"gce_instance\"})\n  )\n)",
```
lyanco, line 986: topk is the wrong function - topk returns the topk points at each timestamp, which can result in a different set of topk points at every timestamp. The right way to filter to topk by mean is to do something like:

```
rate(metric_name[5m])
  and
topk(5, 
  sum_over_time(
    metric_name[3h]
     @ end() offset 1m
))
```

Which picks the top 5 lines based on their average over the last 3 hours. To make this dynamic we would need something like the $__range variable in the FE, but we don't have that.

But honestly we could probably just scrap the top 5; what's the harm really in showing all time series?

mrsillydog: @lyanco Seems reasonable - the simpler approach definitely seems like scrapping the top 5. We've been using `topk(5` pretty heavily throughout these dashboards, should we get rid of it everywhere?

lyanco: Chatted internally - yes please, and also going forward if the dashboard can be represented using ListTimeSeries, then could you convert to that instead? LTS has an automatic top-X built into it which is preferable. If it can't be represented by LTS then PromQL with no topk is good. No need to redo what you've already done (besides remove the topk)
I'll email this to you all as well to keep a paper trail

PR #1022, opened 2025-06-10
PromQL: GKE Enterprise Cluster Kubernetes Events

mrsillydog:
This PR updates the GKE Enterprise Cluster Observability Kubernetes Events Dashboard to use PromQL instead of the deprecated MQL.

MQL:
image

PromQL:
image


mrsillydog:
https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/pull/1062

PR #1023, opened 2025-06-10
PromQL: GKE Enterprise Cluster Memory

mrsillydog:
This PR updates the GKE Enterprise Cluster Observability Memory Dashboard to use PromQL instead of the deprecated MQL.

MQL:
image
image

PromQL:
image
image


mrsillydog:
One thing I am just now noticing is that the `Cluster Unrequested Memory (Top 5 Clusters)` panel is duplicated - this is the case for both MQL and PromQL, so technically this conversion is correct, but I'm thinking it might be worth updating this. Based on one of the other GKE Enterprise dashboards, I'm thinking one of those two panels should be `Cluster Unused Memory (Top 5 Clusters)` (as opposed to `Unrequested`). Whoever reviews this, let me know if you'd like me to make that update.

PR #1028, opened 2025-06-11
PromQL: Redis GCE

braxtonorban:
This PR updates the Redis GCE Overview Dashboard to use PromQL instead of the deprecated MQL.

Before:
image

After: 
image


braxtonorban:
https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/pull/1057

PR #1035, opened 2025-06-12
PromQL: GKE Active/Idle clusters

mrsillydog:
This PR updates the GKE Enterprise Project Observability Memory Dashboard to use PromQL instead of the deprecated MQL.

## Conversion Issues

Windowing - MQL allows you to set the interval of data points for the metric, and PromQL basically just picks an intelligent interval based on the overall window.

Also, the Latest Value column in the table just seems to work different for the two query languages - of the two, PromQL does seem to be more what I'd expect, so that's actually an upgrade. MQL seems to average the datapoints, whereas PromQL actually provides the value of the last reported datapoint.

### Query 1

To attempt to prove this is as close as we get for query 1, here's a couple screenshots of the line chart version of these queries over the same window, but with the MQL query adjusted to replace 7d with 1m (PromQL's default):

MQL:
image

PromQL:
image

### Query 2

To attempt to prove this is as close as we get for query 2, here's a couple screenshots of the line chart version of these queries over the same window, but with the queries adjusted to replace 7d with 1m.

MQL:
image

PromQL:
image

For reference, here's what the exact same query looks like over the last two weeks with the 7d interval back in.

MQL:
image

PromQL:
image

### DIFF
--- a/dashboards/google-kubernetes-engine/gke-active-idle-clusters.json
+++ b/dashboards/google-kubernetes-engine/gke-active-idle-clusters.json
@@ -1,61 +1,72 @@
 {
-  "category": "CUSTOM",
   "displayName": "GKE Active/Idle clusters",
+  "dashboardFilters": [],
+  "labels": {},
   "mosaicLayout": {
-    "columns": 12,
+    "columns": 48,
     "tiles": [
       {
-        "height": 8,
+        "height": 32,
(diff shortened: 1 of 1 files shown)


mrsillydog:
https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/pull/1076

PR #1038, opened 2025-06-13
Update "GKE TPU Node Pool Status" dashboard: make node_pool_name filter multi-select

jakubin:
Developed under b/423634920


erain (approved):
LGTM

PR #1042, opened 2025-06-13
Nginx PromQL Migration Proposal

braxtonorban:
This PR updates the NGINX Overview Dashboard to use PromQL instead of the deprecated MQL.

Before:
image

After:
image

Conversion Issues:

### DIFF
--- a/dashboards/nginx/overview.json
+++ b/dashboards/nginx/overview.json
@@ -176,33 +176,25 @@
         "height": 4,
         "width": 4,
         "widget": {
-          "title": "CPU % Top 5 VMs",
-          "id": "",
+          "title": "VMs CPU %",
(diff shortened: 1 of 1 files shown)


yqlu:
Hi @braxtonorban can you drop dashboards/couchdb/couchdb-gce-overview.png from this diff (it's ok if you force-push to this branch) so that we can merge it cleanly?

braxtonorban:
> Hi @braxtonorban can you drop dashboards/couchdb/couchdb-gce-overview.png from this diff (it's ok if you force-push to this branch) so that we can merge it cleanly?

Hello @yqlu, I just removed this screenshot from this diff. Let me know if you need anything else!

PR #1043, opened 2025-06-13
PromQL: Elasticsearch Node

mrsillydog:
This PR updates the Elasticsearch Node GCE Overview Dashboard to use PromQL instead of the deprecated MQL.

Before:
image
image
image

After:
image
image
image

Conversion Issues:


dashboards/elasticsearch/elasticsearch-node-gce-overview.json
```diff
@@ -391,7 +395,7 @@
                 "plotType": "LINE",
                 "targetAxis": "Y1",
                 "timeSeriesQuery": {
-                  "timeSeriesQueryLanguage": "def top_5_cpu_filtered_by_metric filter_metric =\n  fetch gce_instance\n  | { t_cpu: metric 'compute.googleapis.com/instance/cpu/utilization'\n    ; t_filter_metric: metric $filter_metric }\n  | join\n  | value t_cpu.value.utilization\n  | group_by [resource.project_id, resource.zone, metric.instance_name], 1m,\n      [value_utilization_mean: mean(t_cpu.value.utilization)]\n  | top 5\n  | every 1m;\n\n@top_5_cpu_filtered_by_metric\n  'workload.googleapis.com/elasticsearch.node.http.connections'"
+                  "prometheusQuery": "100 *\navg by (project_id, zone, instance_name) (\n  avg_over_time(\n    compute_googleapis_com:instance_cpu_utilization{monitored_resource=\"gce_instance\"}[1m]\n  )\n  and\n  on(instance_id, project_id, zone)\n  (\n    workload_googleapis_com:elasticsearch_node_http_connections{monitored_resource=\"gce_instance\"}\n  )\n)"
```
yqlu: Does this need a unit override? Like this: https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/pull/1041/commits/9c85dbb1930f4f0b497cf74e4afc8c81ce176c9b

PR #1048, opened 2025-06-13
PromQL: Active Directory DS

mrsillydog:
This PR updates the Active Directory Domain Services GCE Overview Dashboard to use PromQL instead of the deprecated MQL.

Before:
image
image

After:
image
image

Conversion Issues:


yqlu (approved):

mrsillydog:
@yqlu - just letting you know that I don't have write permissions on this repo, so if you could merge any PRs you approve, that would be appreciated. Thanks!

PR #1063, opened 2025-06-13
PromQL: GKE Enterprise Cluster Observability Memory

mrsillydog:
This PR updates the GKE Enterprise Cluster Observability Memory Dashboard to use PromQL instead of the deprecated MQL.

MQL:
image
image

PromQL:
image
image

### DIFF
--- a/dashboards/google-kubernetes-engine-enterprise/gke-enterprise-cluster-observability-memory.json
+++ b/dashboards/google-kubernetes-engine-enterprise/gke-enterprise-cluster-observability-memory.json
@@ -236,63 +236,6 @@
         "height": 16,
         "width": 24,
         "widget": {
-          "title": "Cluster Total Memory (Top 5 Clusters)",
-          "xyChart": {
-            "chartOptions": {
-              "displayHorizontal": false,
-              "mode": "COLOR"
-            },
-            "dataSets": [
-              {
-                "plotType": "LINE",
-                "targetAxis": "Y1",
-                "timeSeriesQuery": {
-                  "prometheusQuery": "label_replace(\n    sum by (location, cluster_name) (\n        kubernetes_io:node_memory_total_bytes{monitored_resource=\"k8s_node\"}\n        or \n        kubernetes_io:anthos_node_memory_total_bytes{monitored_resource=\"k8s_node\"}\n    ),\n    \"cluster_name\", \"$1\", \"cluster_name\", \"(?:.+/|^)(.+)\"\n)",
-                  "unitOverride": "By"
-                }
-              }
-            ],
-            "thresholds": [],
-            "yAxis": {
-              "scale": "LINEAR"
-            }
-          }
-        }
-      },
-      {
-        "yPos": 64,
-        "xPos": 24,
-        "height": 16,
-        "width": 24,
-        "widget": {
-          "title": "Cluster Unrequested Memory",
-          "xyChart": {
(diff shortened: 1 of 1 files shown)


dashboards/google-kubernetes-engine-enterprise/gke-enterprise-cluster-observability-memory.json
```diff
@@ -239,7 +247,8 @@
                 "plotType": "LINE",
                 "targetAxis": "Y1",
                 "timeSeriesQuery": {
-                  "timeSeriesQueryLanguage": "fetch k8s_node\n| { metric kubernetes.io/node/memory/total_bytes\n  ; metric kubernetes.io/anthos/node/memory/total_bytes }\n| union\n| align next_older(2m)\n| every 2m\n| group_by [resource.location, resource.cluster_name], .sum()\n| map\n    update[\n      resource.cluster_name: re_extract(resource.cluster_name, '(?:.+/|^)(.+)')]\n| top 5"
+                  "prometheusQuery": "label_replace(\n    sum by (location, cluster_name) (\n        kubernetes_io:node_memory_total_bytes{monitored_resource=\"k8s_node\"}\n        or \n        kubernetes_io:anthos_node_memory_total_bytes{monitored_resource=\"k8s_node\"}\n    ),\n    \"cluster_name\", \"$1\", \"cluster_name\", \"(?:.+/|^)(.+)\"\n)",
```
varun-c: - Can you update the chart title for this chart to remove the top 5 suffix?
- With the updated title, we will end up with 2 charts with the title "Cluster Total Memory" and 2 with the title "Cluster Unrequested Memory". If they are duplicate, should we delete one? If they are not duplicate, how should we distinguish between the two?

cc: @yqlu

yqlu: This is a good catch for an pre-existing issue with the original dashboard even before we undertook this PromQL migration.

It seems like the original dashboard already had two cards for "Cluster Total Memory (Top 5 Clusters)"

The first has MQL of
```
fetch k8s_node
| { metric kubernetes.io/node/memory/total_bytes
  ; metric kubernetes.io/anthos/node/memory/total_bytes }
| union
| align next_older(2m)
| every 2m
| group_by [resource.location, resource.cluster_name], .sum()
| map
    update[
      resource.cluster_name: re_extract(resource.cluster_name, '(?:.+/|^)(.+)')]
| top 5
```
and the second was
```
{ fetch k8s_node
  | metric kubernetes.io/node/memory/total_bytes
  | align next_older(2m)
  | every 2m
  | group_by [resource.location, resource.cluster_name], .sum()
; fetch k8s_node
  | metric kubernetes.io/anthos/node/memory/total_bytes
  | align next_older(2m)
  | every 2m
  | group_by [resource.location, resource.cluster_name], .sum() }
| union
| map
    update[
      resource.cluster_name: re_extract(resource.cluster_name, '(?:.+/|^)(.+)')]
| top 5
```

These look equivalent (just whether you group by first or whether you union across the two metrics first).

So we should just dedupe and keep one of the two.

mrsillydog: I'll dedupe both of the charts that Varun called out, good catch 👍

varun-c (approved):

PR #1065, opened 2025-06-13
PromQL: GKE Enterprise Namespace Observability CPU

mrsillydog:
This PR updates the GKE Enterprise Namespace Observability CPU Dashboard to use PromQL instead of the deprecated MQL.

MQL:
image

PromQL:
image


pintofloyed:
Is it possible to show the top 5 like in timeSeriesQueryLanguage?

mrsillydog:
> Is it possible to show the top 5 like in timeSeriesQueryLanguage?

We've been dropping the top 5 throughout all of these dashboards - I can't find the original conversation where that decision was made by Lee, but it's intentional.

PR #1069, opened 2025-06-13
Address feedback from vLLM friction log

yqlu:
b/406481466

Short list of changes:
- rearrange textbox and scorecards to grant more horizontal space so titles are clipped less
- add p50 / p95 TPOT to top level scorecard summary
- fix scorecard title "requests / s" should be "num running requests"
- fix chart title "Preemptions count" should be "Preemptions rate"
- eliminate legend templates that are hardcoded and don't respond when global group-bys are applied
- fix issue with generation tokens and prompt tokens were stacked but using different y-axis


dashboards/vllm/vllm-prometheus.json
```diff
@@ -1,1475 +1,1620 @@
...
+                "dimensions": [],
+                "legendTemplate": "",
+                "measures": [],
+                "plotType": "LINE",
                 "targetAxis": "Y1",
-                "breakdowns": []
+                "timeSeriesQuery": {
+                  "outputFullDuration": false,
+                  "prometheusQuery": "sum by (model_name,pod)(avg_over_time(vllm:cpu_cache_usage_perc[${__interval}])) * 100",
+                  "unitOverride": "%"
```
JeffLuoo: nit: Use
```
                  "unitOverride": "10^2.%"
```
so that you don't need to * 100 in the query.

dashboards/vllm/vllm-prometheus.json
```diff
@@ -1,1475 +1,1620 @@
...
-                  "unitOverride": "",
-                  "outputFullDuration": false
-                },
-                "plotType": "STACKED_BAR",
-                "legendTemplate": "Generation tokens",
-                "minAlignmentPeriod": "60s",
-                "targetAxis": "Y1",
-                "breakdowns": []
+                  "outputFullDuration": false,
+                  "prometheusQuery": "sum by (pod,model_name)(avg_over_time(vllm:gpu_cache_usage_perc[${__interval}])) * 100",
```
JeffLuoo: ditto

yqlu: Good idea, done.

PR #1072, opened 2025-06-13
PromQL: GKE Enterprise Project Observability Memory

mrsillydog:
This PR updates the GKE Enterprise Project Observability Memory Dashboard to use PromQL instead of the deprecated MQL.

One thing to note is that I've updated many of the panel titles - this is because a lot of them referenced Cluster, rather than Project. I'm assuming this was a copy-paste mistake, since there is a cluster version of this dashboard as well, and figured it was an easy correction while I was in here.

MQL:
image
image

PromQL:
image
image


varun-c (commented):
Should the first 4 chart titles of this dashboard also be prefixed with "Project " given they are also aggregating the data for the entire project?

An alternative would be to drop the "Project " suffix from all the chart titles, as the dashboard title specifies it, and the chart legends will hopefully make it clear that it is for the projects.

mrsillydog:
Sounds good, I will drop "Project" from the latter 6.

varun-c (approved):

PR #1077, opened 2025-06-13
Vault PromQL Migration Proposal

braxtonorban:
This PR updates the Vault GCE Overview Dashboard to use PromQL instead of the deprecated MQL.


dashboards/vault/vault-gce-overview.json
```diff
@@ -586,33 +578,25 @@
...
-                "measures": [],
                 "plotType": "LINE",
                 "targetAxis": "Y1",
                 "timeSeriesQuery": {
-                  "outputFullDuration": false,
-                  "timeSeriesQueryLanguage": "def top_5_memory_filtered_by_metric filter_metric =\n  fetch gce_instance\n  | { t_memory:\n        metric 'agent.googleapis.com/memory/percent_used'\n        | filter metric.state = 'used'\n    ; t_filter_metric: metric $filter_metric }\n  | join\n  | value val(0)\n  | group_by [metadata.system.name, resource.project_id, resource.zone], 1m,\n      .mean()\n  | top 5\n  | every 1m;\n\n  @top_5_memory_filtered_by_metric 'workload.googleapis.com/vault.core.request.count'",
-                  "unitOverride": ""
+                  "prometheusQuery": "avg by (project_id, zone, instance_id) (\n  avg_over_time(\n    agent_googleapis_com:memory_percent_used{monitored_resource=\"gce_instance\", state=\"used\"}[1m]\n  )\n  and on(project_id, zone, instance_id) (\n    workload_googleapis_com:vault_core_request_count{monitored_resource=\"gce_instance\"}\n  )\n)",
```
yqlu: I think the top level should be `avg by (project_id, zone, instance_name)` instead of `instance_id`? `This is...
- more consistent with the converted CPU chart
- closer to the original MQL one with the legend template
- more human-readable overall.

I just realized I should have been looking out for this but missed it for the PRs I had approved earlier (IIS, Tomcat, Varnish, etc) so if you agree, I think we need a follow-up PR to make this change for all the dashboards that have already been approved. Thanks and sorry for not pointing this out earlier!

braxtonorban: That's a good point that I definitely considered while working on these conversions, although it didn't quite seem to work out in my testing. Here's what @mrsillydog wrote from another PR: "Small problem converting the Memory % Top 5 VMs panel - the legend is different, since we're using instance_id instead of instance_name. The reason for this is because metadata labels appear to not survive joins in PromQL - not sure if that's a quirk of the language itself or just of GCM, but this seemed the best workaround."

After doing another quick test using instance_name in the query on my dashboards, it looks like there's still issues that force us to use instance_id in PromQL. I would love to have the legend be more human readable and better replicate the original query, so if you have any thoughts about how to overcome this gap between MQL and PromQL, please let me know!

image
image
image

yqlu: Thanks. I can verify this behavior.

- If I plot a workload metric by themselves (no aggregation) like `workload_googleapis_com:memcached_bytes{monitored_resource="gce_instance"}`, I see that they have labels `monitored_resource`, `instance_id`, `project_id`, `zone`
- If I plot the agent memory metric by itself (no aggregation) like `agent_googleapis_com:memory_percent_used{monitored_resource="gce_instance", state="used"}`, I see that it has labels `monitored_resource`, `instance_id`, `project_id`, `zone`, `state`

BUT if I switch to the query builder I see that there is a `name` label on the memory metric which I can group by:
Screenshot 2025-07-09 at 4 16 28 PM
And if I convert that query builder to PromQL, I see that I can get access to that label by performing an aggregation over it:
Screenshot 2025-07-09 at 4 18 10 PM
(I am not entirely sure how this works under the hood...)

and now the first half of the query has `project_id`, `instance_id`, `zone`, so I can still do a join with the workload metric on those three labels, while preserving the `metadata_system_name` label:
image

If we wanted to, we could do a final round of aggregation on top of this to strip out the `instance_id` (which was needed for the join, but not needed for the legend), or use a legend template to emphasize the name but not the instance_id, but it's not as critical.

@braxtonorban @mrsillydog please let me know if you can get this approach working across the other GCE agent dashboards -- I see that many of them use the same pattern of joining CPU and memory against the workload metric just to perform a filter down to the applicable VMs. So sorry I didn't point this out before!

braxtonorban: Thank you for the clarification! We'll update the PRs that haven't been reviewed and send in a PR to correct the dashboards that were already merged.

yqlu: Great, appreciate the pivot!

PR #1081, opened 2025-06-16
PromQL: GKE Compute Resources - Node View

mrsillydog:

One caveat is that it seems like the underlying metrics for PromQL and MQL are just different here - while the gke-dev-default-pool-6b815c07-a0th metrics are identical between the screenshots, the gke-dev-default-pool-6b815c07-bsug metrics seem to always be similar, but not identical. I did dig down to the most basic PromQL and MQL queries and confirmed that this difference extends even to the simplest queries, just getting raw metrics.

MQL:
image
image
image

PromQL:
image
image
image


dashboards/google-kubernetes-engine/gke-compute-resources-node-view.json
```diff
@@ -28,744 +28,653 @@
...
         }
       },
       {
-        "xPos": 24,
         "yPos": 24,
-        "width": 24,
         "height": 16,
+        "width": 24,
         "widget": {
+          "title": "CPU Usage per Node - Top 50",
```
johnbryan: can we remove "top 50" from a few titles? It looks like that's no longer true

dashboards/google-kubernetes-engine/gke-compute-resources-node-view.json
```diff
@@ -272,7 +272,7 @@
                 "plotType": "STACKED_AREA",
                 "targetAxis": "Y1",
                 "timeSeriesQuery": {
-                  "timeSeriesQueryLanguage": "{ fetch k8s_node\n  | metric 'kubernetes.io/node/cpu/allocatable_utilization'\n| ${project_id}\n| ${location}\n| ${cluster_name}\n  | group_by 1m,\n      [value_allocatable_utilization_mean: mean(value.allocatable_utilization)]\n  | every 1m\n; fetch k8s_node\n  | metric 'kubernetes.io/node/cpu/allocatable_cores'\n| ${project_id}\n| ${location}\n| ${cluster_name}\n  | group_by 1m, [value_allocatable_core_mean: mean(value.allocatable_cores)]\n  | every 1m }\n| join\n| value [scaled_util: val(0) * val(1)]\n| top 50"
+                  "prometheusQuery": "avg_over_time(kubernetes_io:node_cpu_allocatable_utilization{monitored_resource=\"k8s_node\", ${project_id}, ${cluster_name}, ${location}}[1m])\n*\navg_over_time(kubernetes_io:node_cpu_allocatable_cores{monitored_resource=\"k8s_node\", ${project_id}, ${cluster_name}, ${location}}[1m])"
```
johnbryan: We should generally use `[${__interval}]` instead of `[1m]` in queries. It defaults to 1m, but for longer time ranges like a 1-month view of a dashboard, it expands automatically so the dashboard doesn't try to load minute-resolution data for a whole month.

yqlu: As discussed, once everything else is merged we'll do one big PR to catch up the other dashboards to reflect this change.

PR #1082, opened 2025-06-16
Add Provisioned Throughput Dashboard

katebrea:


dashboards/google-vertex-ai/vertex-ai-provisioned-throughput.json
```diff
@@ -0,0 +1,337 @@
...
+                "measures": [],
+                "minAlignmentPeriod": "60s",
+                "plotType": "LINE",
+                "targetAxis": "Y2",
+                "timeSeriesQuery": {
+                  "outputFullDuration": false,
+                  "timeSeriesFilter": {
+                    "aggregation": {
+                      "alignmentPeriod": "60s",
+                      "crossSeriesReducer": "REDUCE_MEAN",
```
johnbryan, line 195: I'd suggest for this chart either taking the total across models (REDUCE_SUM) or breaking down per model (add resource labels as group-by fields). The current definition takes the mean across PublisherModels which I'm not sure is as meaningful of an aggregation

dashboards/google-vertex-ai/vertex-ai-provisioned-throughput.json
```diff
@@ -0,0 +1,337 @@
...
```
johnbryan, line 219: same here, REDUCE_SUM may be nice to get a total of the per-model limits

dashboards/google-vertex-ai/vertex-ai-provisioned-throughput.json
```diff
@@ -0,0 +1,337 @@
...
+      },
+      {
+        "yPos": 29,
+        "height": 6,
+        "width": 48,
+        "widget": {
+          "title": "",
+          "id": "",
+          "text": {
+            "content": "The Usage by Request Type vs Limit chart shows usage denominated in characters and/or tokens per second broken down by request type. When the traffic is in the \"dedicated\" time series, the traffic was under the limit indicated by GSUs purchased. When it's in the \"shared\" time series, the traffic was classified as pay-as-you-go traffic. When \"spillover\" traffic exists, it indicates that traffic exceeded the limit indicated by the GSU limit and was charged as pay-as-you-go traffic.\n\nTip: Add a filter for a specific model_user_id and location to see usage for a specific Provisioned Throughput allocation.",
```
johnbryan: I like this explanation! A couple optional things you can consider, but your call:
- Could add a link to docs about these concepts
- You can add the `model_user_id` or other filters as predefined filters on the dashboard using "pin filter" to save it to the dashboard json
pin filter

dashboards/google-vertex-ai/vertex-ai-provisioned-throughput.json
```diff
@@ -0,0 +1,337 @@
...
+      },
+      {
+        "yPos": 4,
+        "height": 4,
+        "width": 48,
+        "widget": {
+          "title": "",
+          "id": "",
+          "text": {
+            "content": "The Utilization chart shows Provisioned Throughput utilization as a percentage across models, regardless of whether that model's usage is denominated in tokens or characters.",
```
yqlu: Why do you need two stacked text widgets, one with title but no content, and the second with content but no title?

This seems cleaner:
image

Or we can even just drop the title if the *dashboard* already has a title and this dashboard is relatively small without multiple sections..

dashboards/google-vertex-ai/vertex-ai-provisioned-throughput.json
yqlu: Also I think more markdown (bullet points, using `` for label strings) will help this to avoid appearing as a big paragraph:
Screenshot 2025-06-16 at 2 59 32 PM

dashboards/google-vertex-ai/vertex-ai-provisioned-throughput.json
katebrea: I think I'll drop the title, you're right that the dashboard itself already has a title; I don't need to restate it.

dashboards/google-vertex-ai/vertex-ai-provisioned-throughput.json
katebrea: The aggregation shouldn't matter here because I'm grouping by all of the fields?

dashboards/google-vertex-ai/vertex-ai-provisioned-throughput.json
katebrea: Now I'm grouping by every label, so the aggregation shouldn't matter

dashboards/google-vertex-ai/vertex-ai-provisioned-throughput.json
katebrea: Added the markdown, didn't add the filter- the top chart shouldn't be filtered for best use; the bottom charts should be IMO.

dashboards/google-vertex-ai/vertex-ai-provisioned-throughput.json
yqlu: Thanks, LGTM. Optional: add a [Learn more] to point to the docs for Provisioned Throughput.

We've been doing this for several of the other Vertex AI dashboards since they're all tied to the same integration, so using any one part of Vertex (like training) will auto-populate all of the Vertex AI dashboards.

dashboards/google-vertex-ai/vertex-ai-provisioned-throughput.json
```diff
@@ -0,0 +1,314 @@
+{
+  "displayName": "Provisioned Throughput Usage and Limit",
```
yqlu: I think this should be synced with the `display_name` in metadata.yaml for consistency ("Vertex AI Provisioned Throughput").

dashboards/google-vertex-ai/vertex-ai-provisioned-throughput.json
johnbryan: This chart looks like it's still not grouping by every label?

johnbryan: "Usage by Request Type vs Limit" specifically

PR #1089, opened 2025-06-17
PromQL: GKE Cost Optimization Signals

mrsillydog:
This PR updates the GKE Cost Optimization Signals Dashboard to use PromQL instead of the deprecated MQL.

This one was quite difficult to convert, I don't think any of the panels actually look identical to how they looked pre-conversion. I looked at just the basic metrics, and they seem to just be different between PromQL and MQL - i.e, `fetch <mr> | metric <metric>` would return different values than `<metric>{monitored_resource="<mr>"}`. If anyone has any ideas on what can be done to make these more aligned, happy to try it out.

Another callout is that one of these panels depended on a user label, which don't seem to be queryable in PromQL. Since it was a ratio, I ended up using ListTimeSeries format and splitting it into two separate panels, one for the numerator, one for the denominator.

Screenshots, helpful or useless as they may be:

MQL:
image

PromQL:
image


pintofloyed:
Can the conflicts be resolved?

dashboards/google-kubernetes-engine/gke-optimization-signals.json
```diff
@@ -1,247 +1,276 @@
...
-                  "unitOverride": ""
+                  "prometheusQuery": "sum by (project_id, location, cluster_name) (\n  kubernetes_io:container_memory_request_bytes{\n    monitored_resource=\"k8s_container\",\n    ${project_id},\n    ${location},\n    ${cluster_name}\n  }\n)\n/ \n(\n  sum by (project_id, location, cluster_name) (\n    kubernetes_io:node_memory_allocatable_bytes{\n      monitored_resource=\"k8s_node\",\n      ${project_id},\n      ${location},\n      ${cluster_name}\n    }\n  ) and\n  sum by (project_id, location, cluster_name) (\n    kubernetes_io:node_memory_allocatable_bytes{\n      monitored_resource=\"k8s_node\",\n      ${project_id},\n      ${location},\n      ${cluster_name}\n    }\n  ) > 0\n)\n"
```
pintofloyed: Do you think the denominator from`sum by (ABC) / sum by(XYZ) and sum by(XYZ) > 0` be combined to just `sum by (ABC) / sum by(XYZ) > 0`? I can see the same results with this change for the bottom 4 widgets. Is that the right thing to do? That would make the query a bit cleaner.

mrsillydog: Good catch, I cleaned up the queries in question and added something similar to make other queries in this dashboard safer.

pintofloyed (approved):

PR #1092, opened 2025-06-18
PromQL: GKE System logs

mrsillydog:

One minor caveat on the screenshots - I replaced the 30d time window with 7d, because that window actually allowed the values to relatively match - MQL was very inconsistent with its values based on the window, whereas PromQL strictly increased as the window got larger. This resulted in a substantial mismatch.

MQL:
image

PromQL:
image


pintofloyed (approved):
Can you sync this to head?

pintofloyed:
Need to sync this again before submitting.

PR #1093, opened 2025-06-18
PromQL: Dataflow logs

mrsillydog:

An obvious caveat on the screenshot is that the `Dataflow Byte Count` is blank for MQL and not for PromQL  - this is because of the 1d window on that query. There's only a short span of datapoints, all within an hour or two of each other, and therefore a 1d sampling interval is just too long for MQL to actually show anything. To attempt to add some confidence to the conversion, I've attached screenshots of the same query, but with a 1m interval, over a much smaller window.

MQL (full dashboard):
image

PromQL (full dashboard):
image

MQL (adjusted blank panel):
image

PromQL (same panel):
image


pintofloyed (approved):
Please sync this

PR #1095, opened 2025-06-20
PromQL: VPC Flow Logs Monitoring

mrsillydog:
This PR updates the VPC Flow Logs Monitoring Dashboard to use PromQL instead of the deprecated MQL.

To prove equivalence, here are screenshots of two different versions of the dashboard over the same time period. The former is the previous version of the dashboard, the latter the updated version of the dashboard. Obviously, the panel that has been updated is slightly different due to innate differences between the languages, but this is the closest we get to true parity. If anything, the PromQL is more precise.

MQL (updated panel only):
image

PromQL (same panel):
image


pintofloyed:
Please sync this with HEAD.

PR #1096, opened 2025-06-20
PromQL: HTTP/S LB Backend Services

mrsillydog:
This PR updates the HTTP/S LB Backend Services Dashboard to use PromQL instead of the deprecated MQL.

One small difference that appears to crop up here is that in the first of the two updated panels, there are more time series in the legend - not sure why that is. I think it's just because the time series that are getting dropped from the PromQL only have data for a very very short period of time, maybe less than a minute, which maybe doesn't work with PromQL's increase[1m]? The time series that don't get dropped are graphed identically.

MQL:
image

PromQL: 
image


pintofloyed:
Can you sync this PR?

dashboards/networking/https-lb-backend-services-monitoring.json
```diff
@@ -74,16 +80,19 @@
...
               "timeSeriesQuery": {
-                "timeSeriesQueryLanguage": "fetch https_lb_rule::loadbalancing.googleapis.com/https/backend_request_count\n| { t_0:\n      filter metric.response_code_class = 500\n      | align delta()\n      | group_by [resource.backend_target_name],\n          [value_backend_request_count_aggregate:\n             aggregate(value.backend_request_count)]\n  ; t_1:\n      ident\n      | align delta()\n      | group_by [resource.backend_target_name],\n          [value_backend_request_count_aggregate:\n             aggregate(value.backend_request_count)] }\n| outer_join [0]\n| value\n    [t_0_value_backend_request_count_aggregate_div:\n       div(t_0.value_backend_request_count_aggregate,\n         t_1.value_backend_request_count_aggregate)]"
+                "prometheusQuery": "(\n  sum by (backend_target_name) (\n    increase(loadbalancing_googleapis_com:https_backend_request_count{response_code_class=\"500\",monitored_resource=\"https_lb_rule\",}[1m])\n  )\n  /\n  sum by (backend_target_name) (\n     increase(loadbalancing_googleapis_com:https_backend_request_count{monitored_resource=\"https_lb_rule\",}[1m])\n  )\n)\n"
```
pintofloyed, line 91: Can you use [${__interval}] instead of [1m] in both the promql queries

mrsillydog: This will be getting taken care of in a later PR, per YQ's comment here https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/pull/1081#discussion_r2207881801

For now, we're just going to merge as-is, so that we don't have to update every single one of these PRs individually.

pintofloyed (approved):

PR #1101, opened 2025-06-25
Servicemesh Overview PromQL Migration Proposal

braxtonorban:
This PR updates the Istio - All metrics dashboard to use PromQL instead of the deprecated MQL.

No screenshots are currently available for this conversion do to difficulties with the loadgen. However, only a single query was using MQL, and we can be confident in the conversion due to the simplicity of the query. The reference to the metric in PromQL is certainly correct, and the histogram_quantile function mimics the behavior of the MQL function.


dashboards/servicemesh/overview.json
```diff
@@ -546,32 +546,23 @@
         "width": 6,
         "widget": {
           "title": "Server Response Bytes [MEAN]",
```
johnbryan: can we change this title to "(p95)"

dashboards/servicemesh/overview.json
```diff
@@ -1,875 +1,1239 @@
...
+              "displayHorizontal": false,
               "mode": "COLOR"
             },
             "dataSets": [
               {
                 "plotType": "LINE",
+                "targetAxis": "Y1",
                 "timeSeriesQuery": {
-                  "timeSeriesQueryLanguage": "fetch k8s_container\n| metric 'istio.io/service/server/response_bytes'\n| group_by 1m,\n    [value_response_bytes_percentile: percentile(value.response_bytes, 95)]\n| every 1m"
+                  "prometheusQuery": "histogram_quantile(0.95, \n  rate(istio_io:service_server_response_bytes{monitored_resource=\"k8s_container\"}[1m])\n)"
```
johnbryan: ok I found a project that does have some istio traffic and this didn't quite work... it wants a suffix on the promQL metric 
image


Can you do this for the `timeSeriesQuery` portion of this chart json (builder mode instead of promQL)
```
"timeSeriesQuery": {
      "timeSeriesFilter": {
        "filter": "metric.type=\"istio.io/service/server/response_bytes\" resource.type=\"k8s_container\"",
        "aggregation": {
          "perSeriesAligner": "ALIGN_DELTA",
          "crossSeriesReducer": "REDUCE_PERCENTILE_95",
          "groupByFields": [],
          "alignmentPeriod": "60s"
        }
      }
}
```

braxtonorban: I've updated the title and replaced the PromQL query with this builder mode version!

PR #1102, opened 2025-06-25
PromQL: Microservices (gRPC) Monitoring - GKE

mrsillydog:
This PR updates the Microservices (gRPC) Monitoring - GKE Dashboard to use PromQL instead of the deprecated MQL.

Ended up using a lot of ListTimeSeries for this one - histogram/distribution metrics do not really play that nicely with PromQL, and Heat Maps aren't supported with PromQL at all. Even so, the histogram dashboards (any line chart with an identically-named heat map panel next to it) look slightly different between the old and the new - I can't really account for this change, I'm assuming it's just slightly different handling of the same data.

Another thing to note: the "Client Concurrency" panel did not like my underlying data. This was because of the union on the two time series, one of which dropped a label, the other of which didn't, causing a bad union. I translated it as if they both dropped the label.

MQL:
image
image
image
image

PromQL:
image
image
image
image

### DIFF
--- a/dashboards/google-microservices/microservices-grpc-monitoring-gke.json
+++ b/dashboards/google-microservices/microservices-grpc-monitoring-gke.json
@@ -20,7 +20,8 @@
                 "plotType": "LINE",
                 "targetAxis": "Y1",
                 "timeSeriesQuery": {
-                  "timeSeriesQueryLanguage": "fetch k8s_container\n  | metric 'custom.googleapis.com/opencensus/grpc.io/client/completed_rpcs'\n  | filter (metric.grpc_client_status == 'OK')\n  | align rate(1m)\n  | every 1m\n"
+                  "prometheusQuery": "rate(custom_googleapis_com:opencensus_grpc_io_client_completed_rpcs{monitored_resource=\"k8s_container\", grpc_client_status=\"OK\"}[1m])",
+                  "unitOverride": "1/s"
                 }
               }
             ],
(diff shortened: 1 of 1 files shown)


dashboards/google-microservices/microservices-grpc-monitoring-gke.json
```diff
@@ -258,10 +397,71 @@
...
+                      "crossSeriesReducer": "REDUCE_PERCENTILE_99",
+                      "groupByFields": [],
+                      "perSeriesAligner": "ALIGN_DELTA"
+                    },
+                    "filter": "metric.type=\"custom.googleapis.com/opencensus/grpc.io/client/sent_compressed_message_bytes_per_rpc\" resource.type=\"k8s_container\""
+                  }
+                }
+              },
+              {
+                "legendTemplate": "90.percentile",
```
johnbryan: *95

dashboards/google-microservices/microservices-grpc-monitoring-gke.json
```diff
@@ -552,10 +906,71 @@
...
+                      "crossSeriesReducer": "REDUCE_PERCENTILE_99",
+                      "groupByFields": [],
+                      "perSeriesAligner": "ALIGN_DELTA"
+                    },
+                    "filter": "metric.type=\"custom.googleapis.com/opencensus/grpc.io/server/sent_compressed_message_bytes_per_rpc\" resource.type=\"k8s_container\""
+                  }
+                }
+              },
+              {
+                "legendTemplate": "90 percentile",
```
johnbryan: *95

johnbryan (changes requested):

johnbryan (approved):
approved pending small typos I commented on

PR #1105, opened 2025-07-10
Update VMs Memory % panels

braxtonorban:
This PR corrects the VMs Memory % dashboard panels for dashboards which were previously converted and merged. The update was made based on comments left by @yqlu in [this PR.](https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/pull/1077) Here is an example of an update dashboard panel:
image

### DIFF
--- a/dashboards/aerospike/aerospike-gce-overview.json
+++ b/dashboards/aerospike/aerospike-gce-overview.json
@@ -245,7 +245,7 @@
                 "plotType": "LINE",
                 "targetAxis": "Y1",
                 "timeSeriesQuery": {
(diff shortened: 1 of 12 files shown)


braxtonorban:
> Thanks for the consistency pass!
> 
> From a quick search it looks like the following should also be updated. Do you want to do them in this PR or would these be tackled in a separate follow-up?
> 
> * active-directory
> * elasticsearch-node
> * oracle-db (also has a bunch of topk's we probably want to eliminate)
> * solr

Good catch! I added the updates to these dashboards to this PR. Let me know if any other updates are needed!

yqlu (approved):

PR #1111, opened 2025-07-23
Update GKE Nodes/Pods Cluster View queries

yqlu:
Address b/430479409 -- the pod count scorecard counts raw time series, which may be treating multiple time series from the same pod due to cardinality in metric labels as multiple pods.

This replaces the query builder with count by (labels that serve as a unique identifier for the pod) before performing a sum.

Also reworked some of the table setups -- instead of component_name, we use top_level_controller_name to show a breakdown of pod count per workload.


johnbryan:
I think we actually want `count(count by unique labels)` instead of `sum(count by unique labels)`?

yqlu:
Oops, you're right. If I take away the outer `sum` or `count` we can clearly see what we're trying to aggregate over and that a `count` is right here:

image

Also validated this by checking that after the fix, these four add up to the same number and it matches `kubectl` for that cluster:
image

PR #1113, opened 2025-07-31
PromQL Alerts: Flink

braxtonorban:
This PR updated the Flink alerts to use PromQL instead of the deprecated MQL.

MQL:
image
image

PromQL:
image
image

### DIFF
--- a/alerts/flink/flink-high-jvm-memory-non-heap-usage.v1.json
+++ b/alerts/flink/flink-high-jvm-memory-non-heap-usage.v1.json
@@ -3,24 +3,29 @@
   "documentation": {
     "content": "When the JVM nonheap ratio of nonheap used over max nonheap exceeds a threshold(default is 90%), then application's performance may start to degrade. This alert can trigger when for a jobmanager or a taskmanager. To specify a jobmanager, the resource_type should be jobmanager. To specify a taskmanager, the taskmanager_id filter should target the correct taskmanager instance.",
     "mimeType": "text/markdown"
-},
+  },
   "userLabels": {},
   "conditions": [
     {
       "displayName": "Flink - JVM Heap Memory Near Max",
       "conditionMonitoringQueryLanguage": {
         "duration": "0s",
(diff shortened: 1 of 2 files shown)


cocosheng-g (approved):

alerts/flink/flink-high-jvm-memory-non-heap-usage.v1.json
```diff
@@ -3,24 +3,27 @@
...
       "displayName": "Flink - JVM Heap Memory Near Max",
-      "conditionMonitoringQueryLanguage": {
-        "duration": "0s",
-        "query": "fetch gce_instance\n| {\nt_0: metric 'workload.googleapis.com/flink.jvm.memory.nonheap.used'\n| filter (metric.resource_type == 'jobmanager');\nt_1: metric 'workload.googleapis.com/flink.jvm.memory.nonheap.max'\n| filter (metric.resource_type == 'jobmanager')\n}\n| outer_join 0\n| value t_0.value.used / t_1.value.max\n| condition val() > .9\n| every 5m\n| window 5m",
-        "trigger": {
-          "count": 1
-        }
+      "conditionPrometheusQueryLanguage": {
+        "duration": "300s",
+        "evaluationInterval": "300s",
```
lyanco, line 13: Does 300s duration with 300s evaluation interval actually mean that this alert needs to run for 600s before it fires?

If the alert runs every 5 minutes, then does it need 2 5-minute cycles to have two points over which it's true for 5 minutes?

How should this be written so that it does what it intends - that the signal is > .9 for 5 minutes? do we run it every 60s with a duration of 300s?

lyanco: Conclusion: let's use 60s for evaluation interval and keep 300s for the duration.

When duration is not 0, the promql evaluator uses query_range instead of query_instant to generate multiple points over which the condition must be true for the alert to fire. It sets the step of query_range equal to the evaluation interval.

60s evaluation interval with a 300s duration will only fire if 6 60s points in a row are > .9, which feels like the intent of this alert.

300s evaluation interval with a 300s duration will fire if 2 points 300s apart are .9. So the 60s data could look something like this: 1 0 0 0 0 1; with a 300s evaluation interval that would indeed fire, which feels incorrect, while a 60s evaluation interval will correctly see the 0s in between the 1s and not fire.

PR #1120, opened 2025-08-01
PromQL Alerts: Vault

braxtonorban:
This PR updates Vault alerts to use PromQL instead of the deprecated MQL.

Note that only one of the two alerts could be verified using Metrics Explorer, but that the second alert is extremely similar to the first.

MQL: 
image

PromQL:
image


mrsillydog:
Can be closed, replacement: https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/pull/1160

PR #1121, opened 2025-08-01
PromQL Alerts: Elasticsearch

braxtonorban:
This PR updates an Elasticsearch alert to use PromQL instead of the deprecated MQL.

Note that the alerts is written using the UTF-8 naming style, which was changed after the screenshot was taken.

image


mrsillydog:
Can be closed, replacement: https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/pull/1161

PR #1128, opened 2025-08-01
PromQL Alerts: MongoDB

braxtonorban:
This PR updates some MongoDB alerts to use PromQL instead of the deprecated MQL.

Note that `instance_id` is used instead of `metadata_system_name` because neither metric has a "name" attribute.

MQL:
image

PromQL:
image


mrsillydog:
Can be closed, replacement: https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/pull/1166

PR #1129, opened 2025-08-01
PromQL Alerts: Istio Proxy GKE

braxtonorban:
This PR updates an Istio Proxy GKE alert to use PromQL instead of the deprecated MQL.

Note that the metric has certain attributes that should probably be included in the `sum by`, such as `job`, `instance`, and `namespace`. I left out these aggregations to mimic the MQL behavior.

MQL: 
image

PromQL:
image


mrsillydog:
Can be closed, replacement: https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/pull/1167

PR #1130, opened 2025-08-01
PromQL Alerts: Google Network Intelligence Center

braxtonorban:
This PR updates some Google Network Intelligence Center alerts to use PromQL instead of the deprecated MQL.

Note that in PromQL, packet-loss-high.v1 and packet-loss-high.v2 become identical. This is because the timing change that was made between the two versions is reflected in the alert configuration rather than in the query.

image


mrsillydog:
Can be closed, replacement: https://github.com/GoogleCloudPlatform/monitoring-dashboard-samples/pull/1168

PR #1144, opened 2025-08-06
Update cloud-data-fusion-monitoring.json

SamiSousa:
Fix tile positions to avoid overlap, which prevents dashboards from being copied

### DIFF
--- a/dashboards/google-cloud-data-fusion/cloud-data-fusion-monitoring.json
+++ b/dashboards/google-cloud-data-fusion/cloud-data-fusion-monitoring.json
@@ -473,7 +473,7 @@
       },
       {
         "yPos": 97,
-        "height": 16,
+        "height": 112,
         "width": 48,
         "widget": {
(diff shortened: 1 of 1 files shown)


dashboards/google-cloud-data-fusion/cloud-data-fusion-monitoring.json
```diff
@@ -473,7 +473,7 @@
       },
       {
         "yPos": 97,
-        "height": 107,
+        "height": 16,
```
johnbryan: is this change right? It seems like now the "Pipeline" collapsible group contains fewer widgets than before. Want to check if that's intentional or if this should have height 112 to contain the rest of the dashboard, like before

SamiSousa: Ah my bad, I thought this was just a widget. Let me fix this!

johnbryan (approved):

PR #1147, opened 2025-08-26
Fix autopilot cluster filtering in "GKE Cost Optimization Signals - Bin Packing" dashboard

Kenji-H:
We observed that sometimes the `node name` system metadata, which we use for autopilot cluster filtering, is missing. The availability of the system metadata is high, but we learned that there will be no backfilling of the missing data in case of metadata provider pipeline system outage. 

We decided to resolve the issue on our end by joining the data with another data that has `node name` as resource label. A resource label always exists, so we will not be affected by missing system metadata.

### DIFF
--- a/.github/review-assignment-config.yaml
+++ b/.github/review-assignment-config.yaml
@@ -5,4 +5,5 @@ reviewers:
   - yqlu
   - pintofloyed
   - johnbryan
+  - samisousa
 numberOfReviewers: 1
--- a/dashboards/google-cloudsql/cloudsql-general.json
+++ b/dashboards/google-cloudsql/cloudsql-general.json
@@ -223,9 +223,7 @@
           "title": "Back Up Failure Logs ",
           "logsPanel": {
             "filter": "logName=~\"projects/.*/logs/cloudaudit.googleapis.com%2Factivity\"\nprotoPayload.methodName=\"cloudsql.backupRuns.create\"\nseverity=ERROR",
-            "resourceNames": [
-              "projects/upheld-momentum-328313/locations/global/logScopes/_Default"
-            ]
+            "resourceNames": []
           }
         }
       },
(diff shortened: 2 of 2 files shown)


Kenji-H:
@johnbryan Thank you for the review. 

The reason I switched to use `sum_over_time` from `avg_over_time` is `avg_over_time` sometimes didn't match with the time series graph on the right. I couldn't find out the exact reason, but my guess is the value is weighted based on the time during which a cluster is running or emitted event number in some time slices, and it somehow affects the calculation. After switching to `sum_over_time`, it started showing the more accurate value.

This is an example case where `sum_over_time` returns more accurate value:

The time series graph says the CPU bin packing is around 34%.
Screenshot 2025-09-03 at 17 53 31

Then, using the following `avg_over_time` query shows a higher value for some reason. 
```
round(
    sum(
        sum by (cluster_name) (avg_over_time(kubernetes_io:container_cpu_request_cores{monitored_resource="k8s_container", ${project_id}}[${__interval}]))
        and
        sum by (cluster_name) (avg_over_time(kubernetes_io:node_cpu_allocatable_cores{monitored_resource="k8s_node", node_name!~"gk3-.*", ${project_id}}[${__interval}]))
    )
    /
    sum(avg_over_time(kubernetes_io:node_cpu_allocatable_cores{monitored_resource="k8s_node", node_name!~"gk3-.*", ${project_id}}[${__interval}]))
    * 100,
    0.1
)
```
[
Screenshot 2025-09-03 at 17 53 38
](url)

After changing `avg_over_time` to `sum_over_time`, the gauge value is closer to the one in the time series graph.
```
round(
    sum(
        sum by (cluster_name) (sum_over_time(kubernetes_io:container_cpu_request_cores{monitored_resource="k8s_container", ${project_id}}[${__interval}]))
        and
        sum by (cluster_name) (sum_over_time(kubernetes_io:node_cpu_allocatable_cores{monitored_resource="k8s_node", node_name!~"gk3-.*", ${project_id}}[${__interval}]))
    )
    /
    sum(sum_over_time(kubernetes_io:node_cpu_allocatable_cores{monitored_resource="k8s_node", node_name!~"gk3-.*", ${project_id}}[${__interval}]))
    * 100,
    0.1
)
```
Screenshot 2025-09-03 at 17 54 37

So I want to go with the `sum_over_time` for now.

PR #1151, opened 2025-09-04
PromQL: Change all dashboard references to `[Xm]` to `[${__interval}]`

mrsillydog:
This PR updates all references to specific time intervals within dashboards to use the PromQL `[${__interval}]`

### DIFF
--- a/alerts/oracledb/high-process-utilization.v1.json
+++ b/alerts/oracledb/high-process-utilization.v1.json
@@ -1,26 +1,29 @@
 {
-    "displayName": "Oracle Database - High Process Utilization",
-    "documentation": {
-        "content": "Alert fires when the database is nearing the process limit, which could result in new processes being blocked from starting.",
-        "mimeType": "text/markdown"
-    },
-    "userLabels": {},
-    "conditions": [
-        {
-            "displayName": "Processes near system max",
-            "conditionMonitoringQueryLanguage": {
-                "duration": "0s",
-                "trigger": {
-                    "count": 1
-                },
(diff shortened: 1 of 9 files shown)


dashboards/postgresql/postgresql-gce-overview.json
```diff
@@ -158,7 +158,7 @@
                 "targetAxis": "Y1",
                 "timeSeriesQuery": {
                   "outputFullDuration": false,
-                  "prometheusQuery": "avg_over_time(workload_googleapis_com:postgresql_db_size{monitored_resource=\"gce_instance\"}[1m:1m])",
+                  "prometheusQuery": "avg_over_time(workload_googleapis_com:postgresql_db_size{monitored_resource=\"gce_instance\"}[${__interval}:${__interval}])",
```
mrsillydog: This, and the other few similar options in this file, are the only ones that I'm not 100% sure about, since they're unique. Can't see any reason why this wouldn't be correct though.

scripts/dashboard-importer/src/dashboards/converter/converter_test.ts
```diff
@@ -77,35 +77,35 @@ function testDashboardConversion() {
   assertTitle(tiles[0], 'Test Time Series');
   assertChartPrometheusExpression(
     tiles[0],
-    'rate(node_cpu_seconds_total{mode="idle"}[5m])',
+    'rate(node_cpu_seconds_total{mode="idle"}[${__interval}])',
```
mrsillydog: Also not entirely sure about these Typescript files, should I be limiting my changes here to just the actual dashboard samples? These seemed similarly useful, since they're dashboard-related.

dashboards/postgresql/postgresql-gce-overview.json
lyanco: This should be [${__interval}:] which will set the step of the subquery to the step of the query

scripts/dashboard-importer/src/dashboards/converter/converter_test.ts
stevezease: Yeah, this change should not touch files under scripts/dashboard-importer as it is separate functionality from the sample library dashboards.

PR #1158, opened 2025-09-05
PromQL Alerts: Nginx

mrsillydog:
This PR updates an Nginx alert to use PromQL instead of the deprecated MQL.

MQL:
image

PromQL:
image

### DIFF
--- a/alerts/flink/flink-high-jvm-memory-non-heap-usage.v1.json
+++ b/alerts/flink/flink-high-jvm-memory-non-heap-usage.v1.json
@@ -3,24 +3,27 @@
   "documentation": {
     "content": "When the JVM nonheap ratio of nonheap used over max nonheap exceeds a threshold(default is 90%), then application's performance may start to degrade. This alert can trigger when for a jobmanager or a taskmanager. To specify a jobmanager, the resource_type should be jobmanager. To specify a taskmanager, the taskmanager_id filter should target the correct taskmanager instance.",
     "mimeType": "text/markdown"
-},
+  },
   "userLabels": {},
   "conditions": [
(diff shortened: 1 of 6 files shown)


alerts/nginx/connections-dropped.v1.json
```diff
@@ -8,19 +8,22 @@
...
-      "conditionMonitoringQueryLanguage": {
-        "duration": "60s",
-        "query": "{ \n    t_0: fetch gce_instance::workload.googleapis.com/nginx.connections_accepted\n    ;\n    t_1: fetch gce_instance::workload.googleapis.com/nginx.connections_handled\n}\n| outer_join 0\n| value val(0) - val(1)\n| align rate(5m)\n| condition val() > 0\n",
-        "trigger": {
-          "count": 1
-        }
+      "conditionPrometheusQueryLanguage": {
+        "duration": "0s",
+        "evaluationInterval": "30s",
+        "query": "(\n  rate({\"workload.googleapis.com/nginx.connections_accepted\", monitored_resource=\"gce_instance\"}[5m])\n  -\n  rate({\"workload.googleapis.com/nginx.connections_handled\", monitored_resource=\"gce_instance\"}[5m])\n) > 0"
```
lyanco, line 14: This alert could be flaky because it's possible that the first metric is reported before the second metric, causing the difference to be > 0. 

Let's add "offset 30s" to both of these to ensure that 1 period of data has arrived before we subtract.

lyanco: Actually ignore this - we offset all promql for prometheus metrics by 1m already

PR #1160, opened 2025-09-05
PromQL Alerts: Vault

mrsillydog:
This PR updates Vault alerts to use PromQL instead of the deprecated MQL.

Note that only one of the two alerts could be verified using Metrics Explorer, but that the second alert is extremely similar to the first.

MQL:
image

PromQL:
image


alerts/vault/vault-token-renew-revoke-alert.v1.json
```diff
@@ -1,26 +1,29 @@
 {
-  "displayName": "Vault - Token Renew/Revoke Alert",
+  "displayName": "Vault - Audit Failure Alert",
   "documentation": {
-    "content": "This alert will be triggered when it is taking longer than one second to renew or revoke tokens. This alert will help notify users when there is a potential problem with backend token storage.\n\n",
+    "content": "This alert will be triggered whenever an audit request or audit response failure is greater than 0. This alert will help notify users when their audit device is potentially blocked. If it is blocked, Vault will become unresponsive.\n",
```
lyanco: The metrics being used are workload.googleapis.com/vault.token.renew.time and workload.googleapis.com/vault.token.revoke.time - I think the deleted displayname and content info is correct for this alert

mrsillydog: Oops, good catch, looks like there was some copy-paste issue in the original PR that got continued over here.

alerts/vault/vault-token-renew-revoke-alert.v1.json
```diff
@@ -8,19 +8,22 @@
...
       "displayName": "New condition",
-      "conditionMonitoringQueryLanguage": {
+      "conditionPrometheusQueryLanguage": {
         "duration": "0s",
-        "query": "fetch gce_instance\n| { metric 'workload.googleapis.com/vault.token.renew.time'\n; metric 'workload.googleapis.com/vault.token.revoke.time' }\n| outer_join 0\n| group_by 1m\n| condition val(0) > 1000 || val(1) > 1000",
-        "trigger": {
-          "count": 1
-        }
+        "evaluationInterval": "30s",
+        "query": "rate({\"workload.googleapis.com/vault.token.renew.time\", monitored_resource=\"gce_instance\"}[1m]) > 1000\nor\nrate({\"workload.googleapis.com/vault.token.renew.time\", monitored_resource=\"gce_instance\"}[1m]) > 1000"
```
lyanco: these both say renew.time; one should be revoke.time

lyanco: Also it looks like these are both Gauges, so rate() is an invalid operator: https://cloud.google.com/monitoring/api/metrics_opsagent#opsagent-vault

avg_over_time is more appropriate

lyanco (changes requested):

alerts/vault/vault-token-renew-revoke-alert.v1.json
lyanco: Or no aggregator, that's fine too

PR #1162, opened 2025-09-05
PromQL Alerts: Google GCE

mrsillydog:
This PR updates Google GCE alerts to use PromQL instead of the deprecated MQL.

Note that we included on (reservation_id) in the PromQL versions of the query, which is not a part of the original MQL. This is because without the inclusion of this expression, the PromQL data does not appear; see the screenshots for an example.

image
image

MQL:
image

PromQL:
image

### DIFF
--- a/alerts/google-gce/reservation-utilization-too-high.v1.json
+++ b/alerts/google-gce/reservation-utilization-too-high.v1.json
@@ -1,21 +1,25 @@
 {
-    "displayName": "Reservation - High Utilization",
-    "userLabels": {},
-    "conditions": [
-      {
-        "displayName": "High Reservation Utilization",
-        "conditionMonitoringQueryLanguage": {
-          "duration": "0s",
-          "query": "fetch compute.googleapis.com/Reservation\n|\n{ metric 'compute.googleapis.com/reservation/used'\n| align next_older(5m) | every 5m ;\nmetric 'compute.googleapis.com/reservation/reserved'\n| align next_older(5m) | every 5m\n}\n| ratio\n| condition val() >= 0.9",
-          "trigger": {
-            "count": 1
-          }
-        }
+  "displayName": "Reservation - High Utilization",
+  "userLabels": {},
+  "conditions": [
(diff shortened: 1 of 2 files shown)


alerts/google-gce/reservation-utilization-too-high.v1.json
```diff
@@ -1,21 +1,25 @@
...
-          }
-        }
+  "displayName": "Reservation - High Utilization",
+  "userLabels": {},
+  "conditions": [
+    {
+      "displayName": "High Reservation Utilization",
+      "conditionPrometheusQueryLanguage": {
+        "duration": "0s",
+        "evaluationInterval": "300s",
```
lyanco: These should be swapped - evaluation interval 60s duration 300s
I believe the intention was to only fire if used/reserved is true for 5 minutes. and this metric is emitted every 60s

alerts/google-gce/reservation-utilization-too-low.v1.json
```diff
@@ -1,21 +1,25 @@
...
-        }
+  "displayName": "Reservation - Low Utilization",
+  "userLabels": {},
+  "conditions": [
+    {
+      "displayName": "Low Usage for 20 hours out of 23 hours",
+      "conditionPrometheusQueryLanguage": {
+        "duration": "0s",
+        "evaluationInterval": "300s",
+        "query": "count_over_time(\n  (\n    ({\"compute.googleapis.com/reservation/used\", monitored_resource=\"compute.googleapis.com/Reservation\"}\n     / on (reservation_id)\n     {\"compute.googleapis.com/reservation/reserved\", monitored_resource=\"compute.googleapis.com/Reservation\"}\n    ) <= 0.1\n  )[23h:5m]\n) > (20 * 12) # 20 hours * (12 5 min intervals in hour)"
```
lyanco, line 10: This looks right to me but I'd like joshuapollock@ to take a look

lyanco (changes requested):

PR #1163, opened 2025-09-05
PromQL Alerts: Zookeeper

mrsillydog:
This PR updates a Zookeeper alert to use PromQL instead of the deprecated MQL.

MQL:
image

PromQL:
image

### DIFF
--- a/alerts/zookeeper/low-open-file-descriptors.v1.json
+++ b/alerts/zookeeper/low-open-file-descriptors.v1.json
@@ -8,19 +8,22 @@
   "conditions": [
     {
       "displayName": "New condition",
-      "conditionMonitoringQueryLanguage": {
+      "conditionPrometheusQueryLanguage": {
         "duration": "0s",
-        "query": " fetch gce_instance\n| { metric 'workload.googleapis.com/zookeeper.file_descriptor.limit'\n; metric 'workload.googleapis.com/zookeeper.file_descriptor.open' }\n| outer_join 0\n| group_by 1m\n| sub\n| condition val() < 1000",
-        "trigger": {
-          "count": 1
-        }
+        "evaluationInterval": "30s",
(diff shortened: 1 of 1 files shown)


alerts/zookeeper/low-open-file-descriptors.v1.json
```diff
@@ -8,19 +8,22 @@
...
       "displayName": "New condition",
-      "conditionMonitoringQueryLanguage": {
+      "conditionPrometheusQueryLanguage": {
         "duration": "0s",
-        "query": " fetch gce_instance\n| { metric 'workload.googleapis.com/zookeeper.file_descriptor.limit'\n; metric 'workload.googleapis.com/zookeeper.file_descriptor.open' }\n| outer_join 0\n| group_by 1m\n| sub\n| condition val() < 1000",
-        "trigger": {
-          "count": 1
-        }
+        "evaluationInterval": "30s",
+        "query": "(\n  {\"workload.googleapis.com/zookeeper.file_descriptor.limit\", monitored_resource=\"gce_instance\"}\n  -\n  {\"workload.googleapis.com/zookeeper.file_descriptor.open\", monitored_resource=\"gce_instance\"}\n) < 1000"
```
lyanco, line 14: @yqlu does the group_by 1m in the MQL translate to anything here? Is that a sum() or sum_over_time(metric[1m])?

yqlu: I don't think it's aggregating over the window, it's just picking / aligning one value for every 1m.

From the [docs](https://cloud.google.com/monitoring/mql/reference#group_by-tabop):
>  | group_by   tsid: RowToTSID,   window: lit-WindowDuration,   aggregator: [ RowSetToVal ]
> ...
> If the window argument is given and the tsid argument is not, a group of rows is produced for each time-series identifier and every aligned period output point. The rows in one group are all the rows with the given time-series identifier whose end time falls in a window between the output time and the time that is window duration earlier.
```

PR #1164, opened 2025-09-05
PromQL Alerts: PostgreSQL

mrsillydog:
This PR updates some PostgreSQL alerts to use PromQL instead of the deprecated MQL.

Note that while the condition for high-connection-utilization.v1.json appear differently in MQL and PromQL, they are functionally identical.

MQL:
image

PromQL:
image

### DIFF
--- a/alerts/postgresql/high-connection-utilization.v1.json
+++ b/alerts/postgresql/high-connection-utilization.v1.json
@@ -1,26 +1,29 @@
 {
-    "displayName": "Postgresql - High Connection Utilization",
-    "documentation": {
-        "content": "Alert fires when active connections are near a threshold of 90. Around this point is where the instance may run into connection issues and may start refusing connections.",
-        "mimeType": "text/markdown"
-    },
-    "userLabels": {},
-    "conditions": [
-        {
-            "displayName": "Connections near system max",
-            "conditionMonitoringQueryLanguage": {
-                "duration": "0s",
-                "trigger": {
-                    "count": 1
-                },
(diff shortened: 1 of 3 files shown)


alerts/postgresql/high-cpu-utilization.v1.json
```diff
@@ -1,27 +1,29 @@
...
-    ],
-    "alertStrategy": {
-        "autoClose": "1800s"
-    },
-    "combiner": "OR",
-    "enabled": true,
-    "notificationChannels": []
+  "displayName": "Postgres - High CPU Utilization",
+  "documentation": {
+    "content": "Alerts whenever the CPU utilization goes above 80% which usually indicates the instance's performance is heavily degraded and likely is going to impact applications reliant on postgres.",
```
lyanco: Says 80% but the alert is on .9, so let's make this 90%

alerts/postgresql/high-connection-utilization.v1.json
```diff
@@ -1,26 +1,29 @@
...
+  "documentation": {
+    "content": "Alert fires when active connections are near a threshold of 90. Around this point is where the instance may run into connection issues and may start refusing connections.",
+    "mimeType": "text/markdown"
+  },
+  "userLabels": {},
+  "conditions": [
+    {
+      "displayName": "Connections near system max",
+      "conditionPrometheusQueryLanguage": {
+        "duration": "60s",
```
lyanco: duration to 0s; we can ignore the window 1m

PR #1165, opened 2025-09-05
PromQL Alerts: OracleDB

mrsillydog:
This PR updates some OracleDB alerts to use PromQL instead of the deprecated MQL.

Note that while the condition for the alerts appears differently in MQL and PromQL, they are functionally identical.

MQL:
image

PromQL:
image

### DIFF
--- a/alerts/oracledb/high-session-utilization.v1.json
+++ b/alerts/oracledb/high-session-utilization.v1.json
@@ -1,26 +1,29 @@
 {
-    "displayName": "Oracle Database - High Session Utilization",
-    "documentation": {
-        "content": "Alerts whenever the database is nearing the session limit, which could result in connections being refused.",
-        "mimeType": "text/markdown"
-    },
-    "userLabels": {},
-    "conditions": [
-        {
-            "displayName": "Sessions near system max",
-            "conditionMonitoringQueryLanguage": {
-                "duration": "0s",
-                "trigger": {
-                    "count": 1
-                },
-                "query": "{\n     t_0: fetch gce_instance::workload.googleapis.com/oracle.session.limit ;\n     t_1: fetch gce_instance::workload.googleapis.com/oracle.session.count\n}\n| group_by [resource.instance_id]\n| outer_join 0\n| window 1m\n| condition val(0) * .9 < val(1)\n"
-            }
-        }
-    ],
-    "alertStrategy": {
-        "autoClose": "604800s"
-    },
-    "combiner": "OR",
-    "enabled": true,
-    "notificationChannels": []
+  "displayName": "Oracle Database - High Session Utilization",
+  "documentation": {
+    "content": "Alerts whenever the database is nearing the session limit, which could result in connections being refused.",
+    "mimeType": "text/markdown"
+  },
+  "userLabels": {},
+  "conditions": [
+    {
+      "displayName": "Sessions near system max",
+      "conditionPrometheusQueryLanguage": {
+        "duration": "60s",
+        "evaluationInterval": "30s",
+        "query": "sum by (instance_id) (\n  {\"workload.googleapis.com/oracle.session.count\", monitored_resource=\"gce_instance\"}\n  /\n  {\"workload.googleapis.com/oracle.session.limit\", monitored_resource=\"gce_instance\"}\n) > 0.9"
+      }
+    }
+  ],
+  "alertStrategy": {
+    "autoClose": "604800s",
+    "notificationPrompts": [
+      "OPENED",
+      "CLOSED"
+    ]
+  },
+  "combiner": "OR",
+  "enabled": true,
+  "notificationChannels": [],
+  "severity": "SEVERITY_UNSPECIFIED"
 }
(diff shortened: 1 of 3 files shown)


alerts/oracledb/high-tablespace-utilization.v1.json
```diff
@@ -1,26 +1,29 @@
...
+    "mimeType": "text/markdown"
+  },
+  "userLabels": {},
+  "conditions": [
+    {
+      "displayName": "Tablespace used space near total",
+      "conditionPrometheusQueryLanguage": {
+        "duration": "60s",
+        "evaluationInterval": "30s",
+        "query": "(\n  sum by (instance_id, tablespace_name) (\n    {\"workload.googleapis.com/oracle.tablespace.size\", monitored_resource=\"gce_instance\", state=\"used\"})\n  /\n  (\n    sum by (instance_id, tablespace_name) (\n      {\"workload.googleapis.com/oracle.tablespace.size\", monitored_resource=\"gce_instance\", state=\"used\"})\n    + \n    sum by (instance_id, tablespace_name) (\n      {\"workload.googleapis.com/oracle.tablespace.size\", monitored_resource=\"gce_instance\", state=\"free\"})\n  )\n) > 0.85"
```
lyanco: the screenshots for this chart vs the mql don't look close to the same

mrsillydog: Yeah, I'm discovering that the screenshots here are not very useful, but I think it's fine. This is because we're altering the approach to the alert condition.
- In MQL, we basically just grab the raw values and then do the calculation in the `| condition` (i.e. `| condition (val(0) + val(1)) * .85 < val(0)`. But for graphing purposes, the condition is ignored, so we just end up graphing the raw values.
- In PromQL, we graph the post-calculated value, only if the condition is met.

More specifically, in MQL we graph A and B, then calculate the condition, and in PromQL we graph A / (A + B) (only if it's > 0.85).

Some examples of the boolean logic equivalence:
MQL: (85 + 15) * .85 < 85 = FALSE
PromQL: 85 / (85 + 15) > 0.85 = FALSE

MQL: (86 + 15) * .85 < 86 = TRUE
PromQL: 86 / (86 + 15) > 0.85 = TRUE

MQL: 84 + 15 * .85 < 84 = FALSE
PromQL: 84 / (84 + 15) > 0.85 = FALSE

lyanco: Yeah but your screenshot has the condition commented out in PromQL, so the time series shouldn't be filtered...

mrsillydog: Maybe I'm misunderstanding, but the condition being commented out in PromQL isn't the key - it's the condition being ignored in the MQL.

Here's functionally what's being graphed in the MQL:
```
{
     t_0: fetch gce_instance::workload.googleapis.com/oracle.tablespace.size
| filter (metric.state == 'used') ;
     t_1: fetch gce_instance::workload.googleapis.com/oracle.tablespace.size
| filter (metric.state == 'free')
}
| group_by [resource.instance_id, metric.tablespace_name]
| outer_join 0
| window 1m
// no condition, which includes a significant calculation
```

and here's what's functionally being graphed in the PromQL:
```
(
  sum by (instance_id, tablespace_name) (
    {"workload.googleapis.com/oracle.tablespace.size", monitored_resource="gce_instance", state="used"})
  /
  (
    sum by (instance_id, tablespace_name) (
      {"workload.googleapis.com/oracle.tablespace.size", monitored_resource="gce_instance", state="used"})
    + 
    sum by (instance_id, tablespace_name) (
      {"workload.googleapis.com/oracle.tablespace.size", monitored_resource="gce_instance", state="free"})
  )
)
// much smaller condition, just the > without the calculation
```

These will result in very different graph results, but the addition of the corresponding condition to each results in them being logically identical. Phrased differently, the graphs are different, but so are the conditions, and these differences are complementary.

alerts/oracledb/high-session-utilization.v1.json
```diff
@@ -1,26 +1,29 @@
...
+  "documentation": {
+    "content": "Alerts whenever the database is nearing the session limit, which could result in connections being refused.",
+    "mimeType": "text/markdown"
+  },
+  "userLabels": {},
+  "conditions": [
+    {
+      "displayName": "Sessions near system max",
+      "conditionPrometheusQueryLanguage": {
+        "duration": "60s",
```
lyanco: duration to 0s

alerts/oracledb/high-tablespace-utilization.v1.json
```diff
@@ -1,26 +1,29 @@
...
+  "documentation": {
+    "content": "Alert fires when a tablespace should be resized as it approaches high utilization, before an interruption in service occurs.",
+    "mimeType": "text/markdown"
+  },
+  "userLabels": {},
+  "conditions": [
+    {
+      "displayName": "Tablespace used space near total",
+      "conditionPrometheusQueryLanguage": {
+        "duration": "60s",
```
lyanco: duration to 0s

overall we should just ignore window 1m IMO

lyanco (approved):

PR #1166, opened 2025-09-05
PromQL Alerts: MongoDB

mrsillydog:
This PR updates some MongoDB alerts to use PromQL instead of the deprecated MQL.

Note that instance_id is used instead of metadata_system_name because neither metric has a "name" attribute.

MQL:
image

PromQL:
image


alerts/mongodb/high-disk-utilization.v1.json
```diff
@@ -1,27 +1,29 @@
...
-        }
+  "displayName": "MongoDB - High Disk Utilization",
+  "documentation": {
+    "content": "If Disk Utilization is above 80%, then it is time to look deeper into adding more storage or cleaning up old docs.",
+    "mimeType": "text/markdown"
+  },
+  "userLabels": {},
+  "conditions": [
+    {
+      "displayName": "New condition",
```
lyanco: MongoDB - High Disk Utilization

PR #1167, opened 2025-09-05
PromQL Alerts: Istio Proxy GKE

mrsillydog:
This PR updates an Istio Proxy GKE alert to use PromQL instead of the deprecated MQL.

Note that the metric has certain attributes that should probably be included in the sum by, such as job, instance, and namespace. I left out these aggregations to mimic the MQL behavior.

MQL:
image

PromQL:
image


alerts/istio-proxy-gke/high-request-duration.v1.json
```diff
@@ -1,26 +1,29 @@
...
+  "documentation": {
+    "content": "When the request duration is high, it indicates performance issues with deployed Envoy proxies, network issues, or slow application response times.",
+    "mimeType": "text/markdown"
+  },
+  "userLabels": {},
+  "conditions": [
+    {
+      "displayName": "Prometheus Target - prometheus/istio_request_duration_milliseconds/histogram - 95th percentile",
+      "conditionPrometheusQueryLanguage": {
+        "duration": "300s",
```
lyanco: duration should be 0s otherwise what will happen it we will require a point above the threshold at both the end (now) and the beginning (5m ago) of the duration. given that the rate is [5m] this means the condition would have to be true for 10m, which is not desired

PR #1175, opened 2025-09-05
PromQL Alerts: Zookeeper GKE

mrsillydog:
This PR updates a Zookeeper GKE alert to use PromQL instead of the deprecated MQL.


alerts/zookeeper-gke/low-free-file-descriptors.v1.json
```diff
@@ -8,19 +8,22 @@
   "conditions": [
     {
       "displayName": "New condition",
```
lyanco: Zookeeper - Prometheus - Low Free File Descriptors
probably don't want New condition in there

lyanco (approved):

PR #1176, opened 2025-09-05
PromQL Alerts: MySQL

mrsillydog:
This PR updates MySQL alerts to use PromQL instead of the deprecated MQL.

No screenshots for these ones, but the queries are very simple.

### DIFF
--- a/alerts/mysql/mysql-high-number-of-log-errors.v1.json
+++ b/alerts/mysql/mysql-high-number-of-log-errors.v1.json
@@ -3,24 +3,27 @@
   "documentation": {
     "content": "Log error are a count of logs collected from mysql_error logs with an error level per 5 minute window. The default condition of error per 5 minute window is 5.",
     "mimeType": "text/markdown"
-},
+  },
   "userLabels": {},
   "conditions": [
     {
       "displayName": "logging/user/mysql.error.count",
-      "conditionMonitoringQueryLanguage": {
-        "duration": "0s",
-        "query": "fetch gce_instance\n| metric 'logging.googleapis.com/user/mysql.error.count'\n| window 5m \n| condition val() > 5\n",
-        "trigger": {
-          "count": 1
-        }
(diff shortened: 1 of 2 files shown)


alerts/mysql/mysql-high-number-of-log-errors.v1.json
```diff
@@ -3,24 +3,27 @@
...
-      "conditionMonitoringQueryLanguage": {
-        "duration": "0s",
-        "query": "fetch gce_instance\n| metric 'logging.googleapis.com/user/mysql.error.count'\n| window 5m \n| condition val() > 5\n",
-        "trigger": {
-          "count": 1
-        }
+      "conditionPrometheusQueryLanguage": {
+        "duration": "300s",
+        "evaluationInterval": "30s",
+        "query": "{\"logging.googleapis.com/user/mysql.error.count\", monitored_resource=\"gce_instance\"} > 5"
```
lyanco: I can't tell what the intent of this alert is. It uses a log based metric which is really weird to begin with. But I guess this is supposed to be a gauge metric? But what is a gauge that's a count. It doesn't make any sense. This isn't you, the MQL alert is written bizarrely.

Looking at the content field, I think what this wants is a sum_over_time(metric[5m]) > 5.

The way it's currently written it's that the value of the metric itself has to be over 5 for 5 minutes, which I don't think is intended.

Is this alert linked to some cloud documentation recommending it? If not I think we should just delete this recommended alert. It's really poorly written and requires setting up a LBM first.

mrsillydog: I couldn't find any sort of documentation around this alert, besides the README describing the log-based metric required, and agree that it just generally seems quite ugly all-around. I'll delete.

mrsillydog: The other two alerts in here actually both require LBMs as well. I might just remove this entire directory.

PR #1179, opened 2025-09-05
PromQL Alerts: Google Workload Manager

mrsillydog:
This PR updates Google Workload Manager alerts to use PromQL instead of the deprecated MQL.

No screenshots for this one, but the queries are very simple, as long as you're careful about the boolean logic. Both of these have been slightly altered to be (X or Y) instead of !(X and Y).

### DIFF
--- a/alerts/google-workload-manager/sap-hana-availability-low.v1.json
+++ b/alerts/google-workload-manager/sap-hana-availability-low.v1.json
@@ -8,19 +8,22 @@
   "conditions": [
     {
       "displayName": "SAPHanaAvailabilityCheck",
-      "conditionMonitoringQueryLanguage": {
-        "duration": "0s",
-        "trigger": {
-          "count": 1
-        },
-        "query": "fetch gce_instance\n| { metric 'workload.googleapis.com/sap/hana/availability'\n ; metric 'workload.googleapis.com/sap/mntmode' }\n| join\n| window 1m\n| value if(t_0.value.availability < 1 && t_1.value.mntmode == false(), false(), true())\n| condition val()"
+      "conditionPrometheusQueryLanguage": {
+        "duration": "60s",
+        "evaluationInterval": "30s",
+        "query": "({\"workload.googleapis.com/sap/hana/availability\", monitored_resource=\"gce_instance\"} >= 1)\nor \n({\"workload.googleapis.com/sap/mntmode\", monitored_resource=\"gce_instance\"} != 0)"
       }
     }
   ],
   "alertStrategy": {
-    "autoClose": "604800s"
+    "autoClose": "604800s",
+    "notificationPrompts": [
+      "OPENED",
+      "CLOSED"
+    ]
   },
   "combiner": "OR",
   "enabled": true,
-  "notificationChannels": []
+  "notificationChannels": [],
+  "severity": "SEVERITY_UNSPECIFIED"
 }
--- a/alerts/google-workload-manager/sap-hana-high-availability-low.v1.json
+++ b/alerts/google-workload-manager/sap-hana-high-availability-low.v1.json
@@ -8,19 +8,22 @@
   "conditions": [
     {
       "displayName": "SAPHanaHighAvailabilityCheck",
-      "conditionMonitoringQueryLanguage": {
-        "duration": "0s",
-        "trigger": {
-          "count": 1
-        },
-        "query": "fetch gce_instance\n| { metric 'workload.googleapis.com/sap/hana/ha/availability'\n  ; metric 'workload.googleapis.com/sap/mntmode' }\n| join\n| window 1m\n| value\n    if(t_0.value.availability < 4 && t_1.value.mntmode == false(), false(),\n      true())\n| condition val()"
+      "conditionPrometheusQueryLanguage": {
+        "duration": "60s",
+        "evaluationInterval": "30s",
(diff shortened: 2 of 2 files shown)


alerts/google-workload-manager/sap-hana-availability-low.v1.json
```diff
@@ -8,19 +8,22 @@
...
     {
       "displayName": "SAPHanaAvailabilityCheck",
-      "conditionMonitoringQueryLanguage": {
-        "duration": "0s",
-        "trigger": {
-          "count": 1
-        },
-        "query": "fetch gce_instance\n| { metric 'workload.googleapis.com/sap/hana/availability'\n ; metric 'workload.googleapis.com/sap/mntmode' }\n| join\n| window 1m\n| value if(t_0.value.availability < 1 && t_1.value.mntmode == false(), false(), true())\n| condition val()"
+      "conditionPrometheusQueryLanguage": {
+        "duration": "60s",
```
lyanco: General comment on MQL's confusingness. I haven't really understood what this | window 1m | is in all these gauge ratio queries. I don't think it converts to 60s duration, which is a retest window. I think it's an aligner similar to PromQL's last_over_time that aligns the ratio output point to the 1m, or perhaps aligns the underlying data to take the most recent point over the last 1m. PromQL does an implicit 5m window for gauge metrics so I think it's basically a no-op and we should set duration to 0 for all of these.

lyanco: @yqlu could I get some help here in interpreting this |window 1m| nonsense, it pops up in nearly every gauge ratio alert

yqlu: You are correct -- both in the diagnosis and the recommendation

https://cloud.google.com/monitoring/mql/reference#window-tabop

window   Specifies the window for alignment operations.
Signature:   Table   | window   window: lit-Duration

The window table operation requires an input table that is aligned by an aligner function whose window Duration argument is the same as the window argument given on this window table operation. This is handled in one of the following ways:
- If the input table is not aligned, then an align operation is inserted with an appropriate aligner function for the input table. The Duration argument to the aligner function is given by this window table operation's Duration argument.
- If the table is aligned but the table operation that aligned it does not have a specific alignment window, the window argument from this window table operation is used as the aligner window argument.
- It is an error if the input table is aligned and the table operation that aligned specifies a window Duration that is different from the Duration argument of this window table operation.

lyanco: OK - I think then what we should do is the following:

for standard |window 1m| with no |every 1m| we should just ignore it, set duration to 0s

for |every 5m|window 5m| we should assume the author's intent was to have a 5 minute retest window, so set duration to 300s (and evaluation interval to 60s)

If you agree, I'll go back through and make these comments

alerts/google-workload-manager/sap-hana-availability-low.v1.json
```diff
@@ -8,19 +8,22 @@
...
       "displayName": "SAPHanaAvailabilityCheck",
-      "conditionMonitoringQueryLanguage": {
+      "conditionPrometheusQueryLanguage": {
         "duration": "0s",
-        "trigger": {
-          "count": 1
-        },
-        "query": "fetch gce_instance\n| { metric 'workload.googleapis.com/sap/hana/availability'\n ; metric 'workload.googleapis.com/sap/mntmode' }\n| join\n| window 1m\n| value if(t_0.value.availability < 1 && t_1.value.mntmode == false(), false(), true())\n| condition val()"
+        "evaluationInterval": "30s",
+        "query": "({\"workload.googleapis.com/sap/hana/availability\", monitored_resource=\"gce_instance\"} >= 1)\nor \n({\"workload.googleapis.com/sap/mntmode\", monitored_resource=\"gce_instance\"} != 0)"
```
lyanco: Lol the original MQL query has the condition completely backwards, which is why your faithful inversion of the boolean logic is technically a correct translation but practically nonsense. I knew something seemed squirrely.

The MQL query says "If availability is less than 100% (1) and the system is not in maintenance mode, then do NOT alert, else alert". Obviously they messed that up and meant to write if(condition, true, false) but wrote the exact opposite. Ayiyiyi

Anyhoo we should fix this, IMO should be 

`"({\"workload.googleapis.com/sap/hana/availability\", monitored_resource=\"gce_instance\"} < 1)\nand \n({\"workload.googleapis.com/sap/mntmode\", monitored_resource=\"gce_instance\"} = 0)"`

alerts/google-workload-manager/sap-hana-high-availability-low.v1.json
```diff
@@ -8,19 +8,22 @@
...
       "displayName": "SAPHanaHighAvailabilityCheck",
-      "conditionMonitoringQueryLanguage": {
+      "conditionPrometheusQueryLanguage": {
         "duration": "0s",
-        "trigger": {
-          "count": 1
-        },
-        "query": "fetch gce_instance\n| { metric 'workload.googleapis.com/sap/hana/ha/availability'\n  ; metric 'workload.googleapis.com/sap/mntmode' }\n| join\n| window 1m\n| value\n    if(t_0.value.availability < 4 && t_1.value.mntmode == false(), false(),\n      true())\n| condition val()"
+        "evaluationInterval": "30s",
+        "query": "({\"workload.googleapis.com/sap/hana/ha/availability\", monitored_resource=\"gce_instance\"} >= 4)\nor \n({\"workload.googleapis.com/sap/mntmode\", monitored_resource=\"gce_instance\"} != 0)"
```
lyanco: same thing here - should be availability < 4 and maintenance mode is 0

PR #1180, opened 2025-09-08
PromQL Alerts: Cassandra

mrsillydog:
This PR updates Cassandra alerts to use PromQL instead of the deprecated MQL.


alerts/cassandra/high-error-rate.v1.json
```diff
@@ -1,24 +1,26 @@
...
-                    "count": 1
-                },
-                "query": "{\n    t_0: fetch gce_instance::workload.googleapis.com/cassandra.client.request.count\n    ; \n    t_1: fetch gce_instance::workload.googleapis.com/cassandra.client.request.error.count\n}\n| outer_join 1\n| group_by [metric.status, metric.operation]\n| value val(1) / if(has_value(val(0)), 1.0, val(0))\n| window 1m\n| condition val() > .1"
-            }
+      {
+        "displayName": "Error Rate higher than 10%",
+        "conditionPrometheusQueryLanguage": {
+          "duration": "0s",
+          "evaluationInterval": "60s",
+          "query": "(\n  sum by (status, operation) (\n    rate({\"workload.googleapis.com/cassandra.client.request.error.count\", monitored_resource=\"gce_instance\"}[1m])\n  )\n  /\n  (\n    sum by (status, operation) (\n      rate({\"workload.googleapis.com/cassandra.client.request.count\", monitored_resource=\"gce_instance\"}[1m])\n    )\n    or vector(1)\n  )\n) > 0.1"
```
lyanco: I think this needs to be `or on() vector(1)`

lyanco: But I'm not sure - usually the vector(1) hack is inside of an aggregation function, IDK if it's required inside of a raw parenthetical like you have here

should graph this to be sure

alerts/cassandra/high-error-rate.v1.json
```diff
@@ -1,24 +1,26 @@
...
     "userLabels": {},
     "conditions": [
-        {
-            "displayName": "Error Rate higher than 10%",
-            "conditionMonitoringQueryLanguage": {
-                "duration": "0s",
-                "trigger": {
-                    "count": 1
-                },
-                "query": "{\n    t_0: fetch gce_instance::workload.googleapis.com/cassandra.client.request.count\n    ; \n    t_1: fetch gce_instance::workload.googleapis.com/cassandra.client.request.error.count\n}\n| outer_join 1\n| group_by [metric.status, metric.operation]\n| value val(1) / if(has_value(val(0)), 1.0, val(0))\n| window 1m\n| condition val() > .1"
```
lyanco, line 16: Once again - someone wrote the opposite of what they intended.

if(has_value(val(0)), 1.0, val(0) says "if a value for cassandra.client.request.count exists, then ignore it and use 1, otherwise use cassandra.client.request.count" which is utter nonsense. Like I'm not crazy, https://cloud.google.com/monitoring/mql/reference#if-func is pretty clear that if() operates like it does in excel: "If the first argument is true, then the second argument is returned and if the first argument is false, then the third argument is returned"

So yeah I think someone was trying to avoid dividing by 0 but this is total nonsense. /rant

alerts/cassandra/high-error-rate.v1.json
```diff
@@ -1,24 +1,26 @@
...
-            }
+      {
+        "displayName": "Error Rate higher than 10%",
+        "conditionPrometheusQueryLanguage": {
+          "duration": "0s",
+          "evaluationInterval": "60s",
+          "query": "(\n  sum by (status, operation) (\n    rate({\"workload.googleapis.com/cassandra.client.request.error.count\", monitored_resource=\"gce_instance\"}[1m])\n  )\n  /\n  sum by (status, operation) (\n    rate({\"workload.googleapis.com/cassandra.client.request.count\", monitored_resource=\"gce_instance\"}[1m])\n  ) > 0\n) > 0.1"
```
lyanco: Let's get rid of the > 0. PromQL will look back 5 minutes to pull a point forward to make a rate. While there might be a period where request.count doesn't report data and error.count does, that certainly won't last for more than 5 minutes, which makes the chances that the ratio has a 0/null denominator basically nil.

lyanco (approved):

PR #1181, opened 2025-09-08
PromQL: Nginx ingress

mrsillydog:
This PR updates Nginx Ingress GKE alerts to use PromQL instead of the deprecated MQL.


alerts/nginx-ingress-gke/connections-dropped.v1.json
```diff
@@ -1,26 +1,29 @@
...
+    "mimeType": "text/markdown"
+  },
+  "userLabels": {},
+  "conditions": [
+    {
+      "displayName": "Dropped Connections",
+      "conditionPrometheusQueryLanguage": {
+        "duration": "0s",
+        "evaluationInterval": "30s",
+        "query": "(\n  label_replace(\n    rate({\"nginx_ingress_controller_nginx_process_connections_total\", state=\"accepted\"}[5m]),\n    \"state\",\n    \"\",\n    \"state\",\n    \"\"\n  )\n  -\n  label_replace(\n    rate({\"nginx_ingress_controller_nginx_process_connections_total\", state=\"handled\"}[5m]),\n    \"state\",\n    \"\",\n    \"state\",\n    \"\"\n  )\n) > 0"
```
lyanco: Don't think this is correct... label_replace just obfuscates the `state` field on view, what actually needs to happen is that the two sides need to have their label sets match in order for the - to work correctly, but this should be done instead by either using ignoring(state) on both sides or by doing sum without (state) (rate...) on both sides

lyanco (approved):

PR #1185, opened 2025-09-15
Inference Gateway: Create integration v2 for new metric prefix

JeffLuoo:
Inference Model metrics are renamed to prefix with `inference_objective_`

### DIFF
--- a/dashboards/gateway-api-inference-extension/metadata.yaml
+++ b/dashboards/gateway-api-inference-extension/metadata.yaml
@@ -7,3 +7,13 @@ sample_dashboards:
     related_integrations:
       - id: gateway-api-inference-extension
         platform: GKE
+        version: "1"
+  -
+    category: GKE Inference Gateway
+    id: inference-extension-prometheus-2
+    display_name: GKE Inference Gateway Prometheus Overview
+    description: "This dashboard has charts displaying throughput, latency, cache usage, and other metrics for GKE Inference Gateway. This dashboard requires integration v2 metrics."
+    related_integrations:
+      - id: gateway-api-inference-extension
+        platform: GKE
+        version: "2"
--- a/integrations/gateway-api-inference-extension/prometheus_metadata.yaml
+++ b/integrations/gateway-api-inference-extension/prometheus_metadata.yaml
@@ -1,6 +1,7 @@
 platforms:
   - type: GKE
     launch_stage: GA
+    version: 1
     detections:
       - characteristic_metric:
           metric_type: prometheus.googleapis.com/inference_model_request_duration_seconds/histogram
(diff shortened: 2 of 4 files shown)


dashboards/gateway-api-inference-extension/metadata.yaml
```diff
@@ -2,8 +2,18 @@ sample_dashboards:
...
-    display_name: GKE Inference Gateway Prometheus Overview
+    display_name: GKE Inference Gateway Prometheus Overview (Legacy)
     description: "This dashboard has charts displaying throughput, latency, cache usage, and other metrics for GKE Inference Gateway."
     related_integrations:
       - id: gateway-api-inference-extension
         platform: GKE
+        version: "1"
+  -
+    category: GKE Inference Gateway
+    id: inference-extension-prometheus
```
yqlu: Dashboards need to have a unique ID

dashboards/gateway-api-inference-extension/metadata.yaml
```diff
@@ -2,8 +2,18 @@ sample_dashboards:
   -
     category: GKE Inference Gateway
     id: inference-extension-prometheus
-    display_name: GKE Inference Gateway Prometheus Overview
+    display_name: GKE Inference Gateway Prometheus Overview (Legacy)
```
yqlu: There should be two dashboard JSONs in this repo:
-  inference-extension-prometheus.json with the old metric names, and "(Legacy)" in the name
- inference-extension-prometheus-2.json with the new metric names

dashboards/gateway-api-inference-extension/metadata.yaml
JeffLuoo: Updated the ID for v2 to `inference-extension-prometheus-2` and the dashboard json to `inference-extension-prometheus-2.json`

dashboards/gateway-api-inference-extension/metadata.yaml
JeffLuoo: Updated. The ID is also updated based on comments above, to align with the dashboard json path.

yqlu (approved):
Just one question about rolling out

integrations/gateway-api-inference-extension/prometheus_metadata.yaml
```diff
@@ -45,3 +46,50 @@ platforms:
         kind: GAUGE
         value_type: DOUBLE
     install_documentation_url: https://cloud.google.com/stackdriver/docs/managed-prometheus/exporters/inference-optimized-gateway
+  - type: GKE
+    launch_stage: GA
```
yqlu: Is the library with the new metric prefix already live? Should this be GA or should you use HIDDEN first?

Note that the integration config changes and the new dashboard config will roll out as part of two different binary rollouts, so it's easier to set this as HIDDEN, have both roll out, validate that the new dashboard shows up correctly for the new integration (hidden behind an experiment), and then set this to GA.

JeffLuoo: Thanks for catching it. Updated.

PR #1188, opened 2025-09-30
New vLLM dashboard and integration

yqlu:
See individual commits below -- the first commit sets up the new dashboard config and integration config as a copy of the old one. The second commit showcases the diff between the old and the new config. 

Refer to https://docs.vllm.ai/en/v0.10.2/design/metrics.html for the upstream description of the changes.

Differences between the old and new dashboard:
- replaced time_per_output_token with inter_token_latency
- dropped the `num_requests_swapped` timeseries i n the "Schedule State" chart
- dropped the "best_of param distribution" chart
- dropped "CPU cache usage %"
- replaced "GPU cache usage %" with "KV cache usage %", using the new metric "kv_cache_usage_perc" (functionally replaces the old one)
- within the "Prefix cache hit rate" chart, dropped the `vllm:cpu_prefix_cache_hit_rate` timeseries and replaced the `vllm:gpu_prefix_cache_hit_rate` one with `vllm:prefix_cache_hits_total`

Old dashboard change:
- Fixed the PromQL where it was summing up cache usage % instead of averaging them (if you have two pods with 40% utilization each, you should report 40%, not 80% utilization total).

Integrations:
- set characteristic metric of the old integration to `gpu_cache_usage_perc` (a metric that only existed before)
- set characteristic metric of the new integration to  `kv_cache_usage_perc` (a metric that only exists now)
- updated the metrics offered by the new integration

### DIFF
--- a/dashboards/vllm/metadata.yaml
+++ b/dashboards/vllm/metadata.yaml
@@ -3,7 +3,7 @@ sample_dashboards:
     category: vLLM
     id: vllm-prometheus
     display_name: vLLM Prometheus Overview
-    description: "This dashboard has charts displaying throughput, latency, cache usage, and other metrics for inference."
+    description: "This dashboard supports vLLM v0 metrics and has charts displaying throughput, latency, cache usage, and other metrics for inference."
     related_integrations:
       - id: vllm
         platform: GKE
@@ -12,7 +12,7 @@ sample_dashboards:
     category: vLLM
     id: vllm-prometheus-2
     display_name: vLLM Prometheus Overview
-    description: "This dashboard has charts displaying throughput, latency, cache usage, and other metrics for inference."
+    description: "This dashboard supports vLLM v1 metrics and has charts displaying throughput, latency, cache usage, and other metrics for inference, supporting vLLM v0 metrics."
     related_integrations:
       - id: vllm
         platform: GKE
--- a/dashboards/vllm/vllm-prometheus-2.json
+++ b/dashboards/vllm/vllm-prometheus-2.json
@@ -1079,7 +1079,7 @@
         "height": 16,
         "width": 16,
         "widget": {
-          "title": "Prefix cache hit rate",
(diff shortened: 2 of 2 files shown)


SamiSousa (approved):

dashboards/vllm/vllm-prometheus-2.json
```diff
@@ -0,0 +1,1504 @@
...
+          }
+        }
+      },
+      {
+        "yPos": 89,
+        "xPos": 16,
+        "height": 16,
+        "width": 16,
+        "widget": {
+          "title": "Prefix cache hit rate",
```
JeffLuoo: I can't see it clearly in the screenshot, does it use value 0 ~ 1 or 0 ~ 100 %. I see other charts use 0 ~ 100 %. Can you align on it?

dashboards/vllm/metadata.yaml
```diff
@@ -7,3 +7,13 @@ sample_dashboards:
     related_integrations:
       - id: vllm
         platform: GKE
+        version: "1"
+  -
+    category: vLLM
+    id: vllm-prometheus-2
+    display_name: vLLM Prometheus Overview
+    description: "This dashboard has charts displaying throughput, latency, cache usage, and other metrics for inference."
```
JeffLuoo: nit: mention that the dashboard requires vLLM v1 metrics (which is v2 integration in the current dashboard repo) in the integration.

dashboards/vllm/vllm-prometheus-2.json
yqlu: It's PromQL so by default it's unitless. I've set a unitOverride of `10^2.%` and added (%) to the chart title.

dashboards/vllm/metadata.yaml
yqlu: Updated both the old and new dashboard descriptions to reflect that they support the vllm v0 metrics and vllm v1 metrics respectively.

